Issue #5, opened 2023-10-04
require-error: minimize false positives


Antonboom:
Looks like no difference in
```go
func TestGetDocFileWithLongLine(t *testing.T) {
	fpath := filepath.Join("testdata", "autogen_exclude_long_line.go")
	_, err := getDoc(fpath)
	assert.NoError(t, err) // <---
}
```

Precondition:
1) No assertions more except error
2) Or no operations after assertion at all

More examples:
<details>
<summary>k8s</summary>

```go
	for _, name := range expectedCertificates {
		_, err = pkiutil.TryLoadKeyFromDisk(certDir, name)
		assert.NoErrorf(t, err, "failed to load key file: %s", name)

		_, err = pkiutil.TryLoadCSRFromDisk(certDir, name)
		assert.NoError(t, err, "failed to load CSR file: %s", name)
	}
```

```go
	for _, testCase := range testCases {
		assert.NoError(t, scheme.Convert(testCase.obj1, testCase.obj2, nil))
		assert.NoError(t, scheme.Convert(testCase.obj2, testCase.obj1, nil))
	}
```

```go
	for _, test := range tests {
		err := ValidateAppArmorProfileFormat(test.profile)
		if test.expectValid {
			assert.NoError(t, err, "Profile %s should be valid", test.profile)
		} else {
			assert.Error(t, err, fmt.Sprintf("Profile %s should not be valid", test.profile))
		}
	}
```

```go
		redirect, err := kubelet.GetPortForward(ctx, tc.podName, podNamespace, podUID, portforward.V4Options{})
		if tc.expectError {
			assert.Error(t, err, description)
		} else {
			assert.NoError(t, err, description)
			assert.Equal(t, containertest.FakeHost, redirect.Host, description+": redirect")
		}
```

```go
	for _, testcase := range testcases {
		actual, err := loadDriverDefinition(testcase.filename)
		if testcase.err == "" {
			assert.NoError(t, err, testcase.name)
		} else {
			if assert.Error(t, err, testcase.name) {
				assert.Equal(t, testcase.err, err.Error())
			}
		}
		if err == nil {
			assert.Equal(t, testcase.expected, actual)
		}
	}
```
</details>

<details>

<summary>Grafana</summary>

```go
	t.Cleanup(func() {
		err := resp.Body.Close()
		assert.NoError(t, err)
	})
```

```go
	compareResponses := func(expected *NormalResponse, actual *NormalResponse, compareErr bool) func(t *testing.T) {
[...]

PR #15, opened 2023-10-28
require-error: minimize the number of false positives

Antonboom:
Closes 

https://github.com/Antonboom/testifylint/issues/5


Antonboom:
@scop @rliebz @max-melentyev @AleksSi

Hi, guys!

Thank you for interest to my project.

Could you please run 
```
go install github.com/Antonboom/testifylint@b6f56e0
testifylint --enable require-error ./...
```

And give me feedback about warnings?
Is it ok now?

Antonboom:
@rliebz thank u for such detailed feedback. Appreciated.

Based on my experience `assert` for errors is suitable in rare cases (the most part you described), usually presented in large or _specific_ code bases. Otherwise this make testing debug annoying during panics, and messages from subsequent assertions usually make no sense or you don’t pay attention to them.

For me this is similar to ignoring errors in primary code:
```
v, _ := operation()
```

I made some googling and found support of my position, e.g. [this](https://gitlab.com/gitlab-org/gitlab-pages/-/merge_requests/180) (but this is really an exaggerated example).

I tested `require-error` checker (as linter as well) on several the biggest go projects and in most cases warnings were valid.
The rest false positives really to hard to analyze and I agree with you.

I'll leave the checker enabled by default, waiting for a new wave of feedback from the community.

Also as a result of our discussions I plan to add two new checkers (`graceful-teardown` and `go-require`), I will keep you updated.

Thanks!

PR #34, opened 2023-11-09
Fix "Greater" false positive

dtomcej:
This PR:

- Adds test case demonstrating the false positive

Notes:

Testifylint is returning a false positive for:

```
empty: use assert.NotEmpty (testifylint)
assert.Greater(t, value, 1)
^
```


dtomcej:
This false positive occurs with `v0.2.1` which is used in `golangci-lint`. Using the current version does not result in this issue

Issue #36, opened 2023-11-11
nil-compare: false positives for `assert.Equal(t, nil, value)` (slice, channel)


dolmen:
`assert.Equal(t, nil, []int(nil))` can't be replaced with `assert.Nil(t, []int(nil))`: the behaviour of testify is different, so `nil-compare` should not replace it.
Same issue for `assert.Equal(t, nil, value)` if `value` is of type `chan struct{}`.

See on the Go Playground: https://go.dev/play/p/jUxLF4j-w89

Note: I found this issue when applying `testifylint` on package `github.com/stretchr/testify/mock`: https://github.com/stretchr/testify/blob/db8608ed63f5bd9abdd03f669f5bb80569c114b6/mock/mock_test.go#L765

dolmen:
I wonder if the testsuite lacks a check that really executes the original test and the "fixed" test to verify that the behavior of testify is the same.
@Antonboom If that check exists, could you point me to it?

Antonboom:
@dolmen, hi!

> If that check exists, could you point me to it?
Before I just checked in playground, but from this issue I introduced some code (show you later).

About issue. It looks like `Equal` bug:
```go
func ObjectsAreEqual(expected, actual interface{}) bool {
	if expected == nil || actual == nil {
		return expected == actual               // You cannot do this, because expected and actual interfaces
	}
```

`Equal` is not consistent with Go `==`:
```go
var nilChan chan struct{}
fmt.Println(nilChan == nil)           // true
fmt.Println(any(nilChan) == any(nil)) // false
```

Also it affects `assert.EqualValues(tm, nilChan, nil)`, because I expect this assertion passed, but this failed.

dolmen:
Whether there is a bug in Testify or not (that might be fixed in the future), testifylint should aim to work with Testify as it is now and even as it was in the past, because we can expect that users will apply testifylint on codebases that still run older versions of Testify.

Antonboom:
@dolmen 

> testifylint should aim to work with Testify as it is now

Totally agree, but not in this case. I dove deeper into your example and it's really a misuse of testify itself.
I described details here – https://github.com/stretchr/testify/issues/1524

> and even as it was in the past

It could be difficult, because some checkers use only fresh testify functions and features.
there is like in Go – support of the two last versions of testify 🙁

Thanks for issue! 🤝

PR #39, opened 2023-11-13
error-nil: only fix when the 'expected' arg is nil

dolmen:
`error-nil` now fixes only when the 'expected' argument is nil. The cases where 'expected' and 'actual' are inversed must be left for handling by another checker (especially because `nil` is handled differently by Equal/Exactly/EqualValues).

Also add more comprehensive test cases for Exactly, EqualValues, NotEqualValues.


Antonboom:
Hi, @dolmen!

Thank you for contribution.

> must be left for handling by another checker

Another checkers is not always a reason for this, because they could be disabled by config.
And some linters do extra work in case of obvious mistake, e.g. `len` and
```go
assert.Equal(t, len(arr), 3)    // Reversed but covered by checker.
```

Why `assert.Equal(t, err, nil)` (and similar) is not obvious mistake and should not be covered?

dolmen:
Testify is full of traps if you reverse the `expected` and `actual` arguments. That's why I choose long ago to avoid this package because too often the method names are misleading. That's also why I find your work on this linter of a high value for the Testify's users community.

`assert.Equal` (and family) has a different behavior for this kind of error:
```go
type errorFunc func()

func (errorFunc) Error() string {
	return "error"
}
```
See [on the Go Playground](https://go.dev/play/p/JTOMqU4G1zq):
```go
var err error
assert.Equal(t, err, nil)
assert.Equal(t, nil, err)
assert.NoError(t, err)

err = errors.New("error1")
assert.NotEqual(t, err, nil)
assert.NotEqual(t, nil, err)
assert.Error(t, err)

err = errorFunc(nil)
assert.NotEqual(t, err, nil)
assert.NotEqual(t, nil, err)
assert.Error(t, err)
```

Antonboom:
> However assert.Equal (and family) has a different behavior for this kind of error:

Anyway linter suggest to replace it with `Error`/`NoError` that works correct (based on plyaground).
No issues here?

> I had a deep look at assert's code, and I found no expected vs actual trap for the nil case

In this case, could you rollback related changes?

> because nil is handled specially
Please, take a look at https://github.com/Antonboom/testifylint/issues/36#issuecomment-1809800423

README.md
```diff
@@ -180,8 +180,11 @@ logic, but without autofix.
 ```go
 ❌   assert.Nil(t, err)
      assert.NotNil(t, err)
-     assert.Equal(t, err, nil)
-     assert.NotEqual(t, err, nil)
+     assert.Equal(t, nil, err)
```
Antonboom, line 183: I tried to make examples as short as possible, because a lot of combinations here.
Grep `// And other variations` for examples.

If you found such changes useful I can add `Exactly` and `EqualValues` to the other checkers description too

Antonboom: Introduced in https://github.com/Antonboom/testifylint/pull/49

Antonboom:
@dolmen 

> However assert.Equal (and family) has a different behavior for this kind of error:

IMO, your snippet is another yet example of misusage of **testify**.
`Equal`-based assertions could not be used with functions (according to `validateEqualArgs`).

Maybe it also related to https://github.com/stretchr/testify/issues/1524

Thanks, but closed 🤝

PR #51, opened 2024-01-12
organize dev tools installation

mmorel-35:
This provides a Makefile to setup dev tools with only one target:
```bash
make dev/tools
```

It removes goimports (duplication with gci) and gofmt (superseded by gofumpt)

It also removes task fmt as golangci-lint--fix is already doing the same job.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,9 +5,9 @@
 ### 1) Install developer tools
 
 ```bash
-# https://taskfile.dev/installation/
-$ go install github.com/go-task/task/v3/cmd/task@latest
-$ task tools:install
+$ make dev/tools
+go install github.com/go-task/task/v3/cmd/task@latest
+task tools:install
 Install dev tools...
 ```
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,16 @@
+GOBIN= ${GOPATH}/bin
+
+# https://taskfile.dev/installation/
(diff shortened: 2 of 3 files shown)


Antonboom:
@mmorel-35, hi!

> This provides a Makefile to setup dev tools with only one target:

I don't see the point in having multiple `*file`. `task` installation is not a big deal.
And `Makefile` introduces inconsistency in workspace.

> It removes goimports (duplication with gci) and gofmt (superseded by gofumpt)

Accepted ([commit](https://github.com/Antonboom/testifylint/commit/4ca7d527258c5f1fa85d816d7b43bddeb42c8332)).

> It also removes task fmt as golangci-lint--fix is already doing the same job.

Not accepted. The current approach allows

1. Separate `fmt` and `lint` steps. E.g. you can do formatting every file saving, but linting (as more heavy task) every commit.
2. Remove `--fix` if there is some issues with it.
3. Do not depend from `golangci-lint` in part of formatting. The bug fix in the separate tool will be available in the linter in several months after.

> go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest

Also not the best practise. Read [more](https://golangci-lint.run/usage/install/#install-from-source).

---

Appreciated your attention, but please focus your efforts on making more meaningful changes:
https://github.com/Antonboom/testifylint/blob/master/CONTRIBUTING.md#open-for-contribution

PR #52, opened 2024-01-12
setup dependabot config for github-actions

mmorel-35:
This also provides hash for github-actions version and read/write permissions for workflow and jobs.

Don’t worry dependabot is able to handle the hash and the commented version.

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>


.github/dependabot.yml
```diff
@@ -1,7 +1,11 @@
 version: 2
 
 updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
```
Antonboom, line 7:
```suggestion
      interval: monthly

```

.github/workflows/ci.yml
```diff
@@ -11,39 +11,48 @@ env:
   GO_VERSION: ^1.21
   GOLANGCI_LINT_VERSION: v1.54.0
 
+permissions:
```
Antonboom, line 14: 1. why not `permissions: read-all`?
2. what is profit from `permissions ` in specific context of this project?

.github/workflows/ci.yml
```diff
@@ -11,39 +11,48 @@ env:
...
+  contents: read
+
 jobs:
   go_install:
     if: github.repository == 'Antonboom/testifylint'
     runs-on: ubuntu-latest
     steps:
-      - uses: rlespinasse/github-slug-action@v4.4.1
-      - uses: actions/setup-go@v4
+      - uses: rlespinasse/github-slug-action@102b1a064a9b145e56556e22b18b19c624538d94 # v4.4.1
```
Antonboom, line 22: I don't like hashes and interested only in major version if we cannot write `@v4.4.1`
is it possible to configure `dependabot` to watch only for major updates?

Antonboom (changes requested):

.github/workflows/ci.yml
mmorel-35: read-all allows way more access than just contents: read.
It's usually better to accord only required permission to actions

.github/workflows/ci.yml
mmorel-35: Having hashes is recommended by OSSF best practices concerning dependencies including github-actions. 
It's of course possible to configure dependabot to watch only major updates

Antonboom: what is profit from it in specific context of this project?

> It's of course possible to configure dependabot to watch only major updates

please do it or close PR since I don't see a threat model here.
looks like change for the sake of change

Antonboom (changes requested):

.github/dependabot.yml
```diff
@@ -1,7 +1,13 @@
 version: 2
 
 updates:
+  - package-ecosystem: github-actions
+    directory: /
+    update-types: semver-major
```
mmorel-35, line 6: Only major updates for github-actions here

Antonboom (approved):

Antonboom:
@mmorel-35 you didn't test your changes :(

Screenshot 2024-01-24 at 08 12 36

please do not use my projects as field for experiments

PR #53, opened 2024-01-12
Remove Testify V2 notice on README as it wont happen

hendrywiranto:
# Background
As mentioned in this announcement [link](https://github.com/stretchr/testify/issues/1089#issuecomment-1812734472) from Testify maintainers

> So, as the current only active maintainer, I'm declaring that v2 will never happen. Or at least a v2 of the github.com/stretchr/testify module with such major breaking changes.

# Description
Remove Testify V2 notice on README as it won't happen

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,13 +475,3 @@ require.NoError(t, err)
 ```
 
 Please [contribute](./CONTRIBUTING.md) if you have ideas on how to make this better.
-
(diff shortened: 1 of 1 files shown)


hendrywiranto:
also, like my other PR on testify, if you want to link the manifesto instead just let me know! Thanks 😁

PR #77, opened 2024-04-07
Add links in CONTRIBUTING.md

ccoVeille:
This would help people to find explanation or reuse an existing issue when implementing it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -178,7 +178,8 @@ Describe a new checker in [checkers section](./README.md#checkers).
 
 **Autofix**: false. <br>
 **Enabled by default**: true. <br>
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -299,6 +301,26 @@ And similar idea for `assert.InEpsilonSlice` / `assert.InDeltaSlice`.
 
 ---
 
+### if-err-fatal
```
Antonboom, line 304: Hi!

1. I meant a more common solution than just fatal replacer.
2. `testifier` proposed to be a separate tool and maybe in separate repo

Please remove this section

Antonboom (changes requested):

CONTRIBUTING.md
ccoVeille: I removed the commit

PR #78, opened 2024-04-10
add new checker: negative-positive

### LINKED ISSUE #76: feature: Use positive/negative instead of assert.Greater(t, 0, a)/assert.Less(t, 0, a)
For records, I started working on the implementation of this feature request.

https://github.com/Antonboom/testifylint/blob/master/CONTRIBUTING.md#negative-positive

I created the issue to keep a track of it, and allow other people to discuss it.

Stay tuned, and ping me if you see no progress in the next month.

ccoVeille:
closes #76


internal/testgen/main.go
```diff
@@ -27,6 +27,7 @@ var checkerTestsGenerators = []CheckerTestsGenerator{
 	BoolCompareTestsGenerator{},
 	ComparesTestsGenerator{},
 	EmptyTestsGenerator{},
+	NegativePostiveTestsGenerator{},
```
Antonboom, line 30: please keep alphabetic order

internal/checkers/checkers_registry.go
```diff
@@ -12,6 +12,7 @@ var registry = checkersRegistry{
 	{factory: asCheckerFactory(NewEmpty), enabledByDefault: true},
 	{factory: asCheckerFactory(NewLen), enabledByDefault: true},
 	{factory: asCheckerFactory(NewCompares), enabledByDefault: true},
+	{factory: asCheckerFactory(NewNegativePositive), enabledByDefault: true},
```
Antonboom, line 15: looks like it should be before Compares

internal/testgen/gen_negative_positive.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return struct {
+		CheckerName       CheckerName
+		InvalidAssertions []Assertion
+		ValidAssertions   []Assertion
+		IgnoredAssertions []Assertion
+	}{
+		CheckerName: CheckerName(checker),
+		InvalidAssertions: []Assertion{
+			{Fn: "Greater", Argsf: "0, a", ReportMsgf: report, ProposedFn: "Positive", ProposedArgsf: "a"},
```
Antonboom, line 30: please, support cases

```
True(a > 0)
True(0 > a)
False(a < 0)
False(0 < a)
```
(`compares` can be disabled and I allow some overlap in the linters)

+ similar ignored assertions

internal/checkers/negative_postive.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"golang.org/x/tools/go/analysis"
+
+	"github.com/Antonboom/testifylint/internal/analysisutil"
+)
+
+// NegativePositive detects situations like
+//
+//	assert.Greater(t, 0, value)
```
Antonboom, line 11: will be updated with new cases (see below) + README

internal/checkers/negative_postive.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// this variable will help us to consider which argument is zero
+	secondArgZero := isZero(call.Args[1])
+	if !xor(isZero(call.Args[0]), secondArgZero) {
+		// this checker only needed if one of the argument is zero
+		// if both are zeros it will be reported by useless-assert checker
+		return nil
+	}
+
+	// ok now we know one of the argument is zero, but not both
```
Antonboom, line 39: please, remove useless comments
code must be self-documented and in ideal picture do not contain comments at all

if you see the place for comment use `// NOTE(author): comment` notation

Antonboom (changes requested):
Thank you!

ccoVeille:
Thanks for the review. I'll try to sort this out in the next few days.

internal/testgen/gen_negative_positive.go
ccoVeille: Good catch. I didn't think about these ones.

internal/checkers/negative_postive.go
ccoVeille: You only refer to these right ?

https://github.com/Antonboom/testifylint/pull/78#discussion_r1561432681

internal/checkers/negative_postive.go
ccoVeille: That's subjective and open to debate, but I will respect your guidance.

internal/checkers/checkers_registry.go
ccoVeille: Indeed

ccoVeille: I'm no longer sure now

Please look at https://github.com/Antonboom/testifylint/pull/78#discussion_r1561432681

internal/testgen/gen_negative_positive.go
ccoVeille: I started working on this, but then it leads to a very complex code in negative_positive.go

It has to replicate a lot of what `compares.go` does

I remember now why I initially stated to put negative-positive after compares.go

It's because compares would request to transform `True(t, a>0)` into `Greater(t, 0, a)`

From my perspective, and because you explained it well here, it should be the way to do.

https://github.com/Antonboom/testifylint?tab=readme-ov-file#chain-of-warnings

I mean compare is intended to detecte the True/False+boolean

https://github.com/Antonboom/testifylint?tab=readme-ov-file#compares

While negative-compare would detect

https://github.com/Antonboom/testifylint/blob/master/CONTRIBUTING.md#negative-positive

I don't think it's a good idea to mix everything.

Antonboom: Please, provide an example of "very complex code". Compares is pretty simple and we need just of part of it.

How I wrote above 

> compares can be disabled and I allow some overlap in the linters

Need to find a balance. If implementation is really complex, then I can agree 👍 

"Chain of warning" is more about full independent linters. But in this case I feel some false negatives.

ccoVeille: I understand you disagree, but here is my point of view:
- compares is about catch True(a > b) and co to use Greater (and co)
- negative-positive is about detecting Greater(a, 0) to use Positive(a) and co

I think our visions differ, I dislike that, but it's OK.

Now, that is said, I implemented the feature that was requested and described here. 

https://github.com/Antonboom/testifylint/blob/7346760ab41bdd2fe72302620199d89725e0b8ed/CONTRIBUTING.md#negative-positive

We can argue and iterate on the need to add `True(a > 0)` (and co) detection to `negative-positive.go`
I have no problem in bringing a new feature such as this one in negative-positive.go, but I would prefer to do it in another PR, with a dedicated issue.

The reason is that bringing positive-negative now seems to me to be a better option for people using your tool.
So can you consider reviewing current code and merging it as-is 

For records, I'm more interested in adding other feature such as zero or error-contains than adding this edge case that is already reported by True(a >0), that will be reported to use Positive(a) on next run.

Antonboom: Hi, @ccoVeille 

thank you for reply
I found some bugs in your PR and want to contribute and test it again on my own – to not waste your time more.

> I'm more interested in adding other feature

welcome ❤️ 

I returned to this PR soon and will describe changelog

ccoVeille: Come back to me when you can. I'm starting working again on "zero" checker

README.md
```diff
@@ -376,6 +378,33 @@ P.S. Related `testify`'s [thread](https://github.com/stretchr/testify/issues/772
 
 ---
 
+### negative-positive
+
+```go
+❌   assert.Greater(t, a, 0)
```
Antonboom, line 384: 1. All cases in README
2. docs is consistent with tests and checker type godoc

internal/checkers/negative_postive.go
```diff
@@ -0,0 +1,200 @@
...
+import (
+	"go/ast"
+	"go/token"
+
+	"golang.org/x/tools/go/analysis"
+
+	"github.com/Antonboom/testifylint/internal/analysisutil"
+)
+
+// NegativePositive detects situations like
```
Antonboom, line 12: All cases in docs.

internal/checkers/negative_postive.go
```diff
@@ -0,0 +1,200 @@
...
+//
+//	assert.Positive(t, value)
+//	assert.Negative(t, value)
+type NegativePositive struct{}
+
+// NewNegativePositive constructs NegativePositive checker.
+func NewNegativePositive() NegativePositive { return NegativePositive{} }
+func (NegativePositive) Name() string       { return "negative-positive" }
+
+func (checker NegativePositive) Check(pass *analysis.Pass, call *CallMeta) *analysis.Diagnostic {
```
Antonboom, line 38: implementation is consistent with `Empty` checker

internal/checkers/negative_postive.go
```diff
@@ -0,0 +1,200 @@
...
+
+		survivingArg, ok := anyVal([]bool{ok1, ok2}, a, b)
+		if ok {
+			return newUsePositiveDiagnostic(expr.Pos(), expr.End(), survivingArg)
+		}
+	}
+	return nil
+}
+
+func isStrictComparisonWith(
```
Antonboom, line 173: analogue of neighbour `isComparisonWith`

internal/testgen/gen_negative_positive.go
```diff
@@ -0,0 +1,232 @@
+package main
+
+import (
+	"strings"
+	"text/template"
+
+	"github.com/Antonboom/testifylint/internal/checkers"
+)
+
+type NegativePositiveTestsGenerator struct{}
```
Antonboom, line 10: Fix typo

internal/testgen/gen_negative_positive.go
```diff
@@ -0,0 +1,232 @@
...
+			{Fn: "False", Argsf: "0 >= a", ReportMsgf: report, ProposedFn: "Positive", ProposedArgsf: "a"},
+
+			{Fn: "Less", Argsf: "a, 0", ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+			{Fn: "Greater", Argsf: "0, a", ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+			{Fn: "True", Argsf: "a < 0", ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+			{Fn: "True", Argsf: "0 > a", ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+			{Fn: "False", Argsf: "a >= 0", ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+			{Fn: "False", Argsf: "0 <= a", ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+		},
+		ValidAssertions: []Assertion{
```
Antonboom, line 44: Valid assertions implies "target" assertion (after replacements).
Just to check that linter do not warn on their our suggestions.

internal/testgen/gen_negative_positive.go
```diff
@@ -0,0 +1,232 @@
...
+			{Fn: "NotEqual", Argsf: "a, a"},
+			{Fn: "Greater", Argsf: "0, 0"},
+			{Fn: "Greater", Argsf: "a, a"},
+			{Fn: "GreaterOrEqual", Argsf: "0, 0"},
+			{Fn: "GreaterOrEqual", Argsf: "a, a"},
+			{Fn: "Less", Argsf: "0, 0"},
+			{Fn: "Less", Argsf: "a, a"},
+			{Fn: "LessOrEqual", Argsf: "0, 0"},
+			{Fn: "LessOrEqual", Argsf: "a, a"},
+			{Fn: "True", Argsf: "a > a"},
```
Antonboom, line 170: https://github.com/Antonboom/testifylint/issues/80

internal/testgen/gen_negative_positive.go
```diff
@@ -0,0 +1,232 @@
...
+			{Fn: "True", Argsf: "a < 0", ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+			{Fn: "True", Argsf: "0 > a", ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+			{Fn: "False", Argsf: "a >= 0", ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+			{Fn: "False", Argsf: "0 <= a", ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+		},
+		ValidAssertions: []Assertion{
+			{Fn: "Positive", Argsf: "a"},
+			{Fn: "Negative", Argsf: "a"},
+		},
+		IgnoredAssertions: []Assertion{
```
Antonboom, line 48: significantly more tests
check the border cases -1, 0, 1 (I like your idea) for any comparison

analyzer/testdata/src/debug/negative_positive_replacements_test.go
```diff
@@ -0,0 +1,56 @@
+package debug
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNegativeReplacements(t *testing.T) {
```
Antonboom, line 10: your replacement 

```go
{Fn: "Greater", Argsf: "0, a", ReportMsgf: report, ProposedFn: "Positive", ProposedArgsf: "a"},
```

is invalid because of arguments order

```go
// Greater asserts that the first element is greater than the second
//
//	assert.Greater(t, 2, 1)
```

But it is difficult to explore, because for `a == 0` both options (`Greater(a,0)` and `Positive(a)`) return `false`.

I fixed it and wrote replacement test to debug and be sure that checker suggestions is right.

Antonboom (commented):
@ccoVeille please review it accurately.

thanks :)

ccoVeille:
Thanks for coding the things you wanted to add.

internal/testgen/gen_negative_positive.go
ccoVeille: > Please, provide an example of "very complex code". Compares is pretty simple and we need just of part of it.
> 
> How I wrote above 
> 
> > compares can be disabled and I allow some overlap in the linters
> 
> Need to find a balance. If implementation is really complex, then I can agree 👍 
> 
> "Chain of warning" is more about full independent linters. But in this case I feel some false negatives.

The code you added via 21776049f952ac584f9603b309ffa8f1edc1d0fe is great, but I would dare to say it's more complex than the code I wrote 😅

I will review the changes thanks

internal/checkers/negative_postive.go
ccoVeille: I like be the functions like this, like the isNotZero and co.

The code is now starting to use function from other checkers, it's OK. But, I would like to suggest an improvement.

What do you think about them moving out of each checker into a "lib" in the same folder ?

So they could be more visible, and would help people to contribute. *Also they could be tested*.

I mean yesterday, I started working on something for checking require.TestingT, the I found later an existing function doing it.

TL;D; I'm OK with current PR, I'm talking about moving then in a new issue/PR

internal/checkers/negative_postive.go
ccoVeille: Can you point out what you meant about this ? I'm unsure what you meant

internal/testgen/gen_negative_positive.go
ccoVeille: Thanks, I feel ashamed, haha

ccoVeille: BTW @Antonboom , an idea for another issue/PR, would you like me to help you in setting up code spell GitHub action for checking typos in the code and documentation ?

internal/testgen/gen_negative_positive.go
ccoVeille: I saw you created it, thanks

analyzer/testdata/src/debug/negative_positive_replacements_test.go
ccoVeille: Oh, thanks for noticing it.

ccoVeille:
I checked up quickly the changes you pointed in your comments.

I replied to some of them.

I will review the changes you made in details later.

internal/checkers/negative_postive.go
ccoVeille: @Antonboom I will resolve this one, but I would like to get your point of view about the idea of moving methods like this into a common file, so xor, isZero, isNotZero and co and are not spread among all files and added where they are needed the first time.

I'm asking because I wrote a few code before noticing they were already available in another file, with a different name than the one I chose

ccoVeille (commented):
Everything is fine to me

I downloaded and launched it on my code base. Everything works as expected.

internal/checkers/negative_postive.go
Antonboom: I meant a code structure: help methods like checker.checkNegative/Positive, switches inside, etc.

internal/testgen/gen_negative_positive.go
Antonboom: I am using GoLand that highlights any issues.
Also `misspell` linter is already enabled, maybe need to configure it properly 🤷‍♂️ 

But I dont want to introduce new tools and tries to keep the project as simple as possible (but yep, this is already monstrous application).

internal/checkers/negative_postive.go
Antonboom: > not spread among all files and added where they are needed the first time.

yep, I followed this approach while code base was not so significant 

1. I did not want to move helpers in separate package, because the most of them requires `Pass` and could not be tested separately from analyzer. 
2. Also I am afraid of "Helpers Anti-Pattern" 

> of moving methods like this into a common file

looks like compromise, something like
- `checkers/helpers_len.go`
- `checkers/helpers_bool.go`
- `checkers/helpers_error.go`
?

I can’t figure out what’s better and that’s why I left everything as it is.

Therefore, I’m open to any proposals, just take into account the points above 👍

ccoVeille: Yes, I was talking about having them in the same package than now.

I like your idea of spitting them by context, so contributor won't be lost in a huge "helpers.go"

ccoVeille: I opened https://github.com/Antonboom/testifylint/issues/84 to continue the discussion there.

internal/checkers/negative_postive.go
ccoVeille: Thanks, I now get it. Yes, it's clearer and consistent.

ccoVeille:
Thanks for your help and review.

I'm happy to have contributed to the tool.

Issue #84, opened 2024-04-20
suggestion: split test helpers in to separate files in same package


ccoVeille:
Right, now the helpers widely spread among the checker files and added where they are needed the first time.

Here is maintainer point of view
> I followed this approach while code base was not so significant 
> They could be reordered:
> 1. I did not want to move helpers in separate package, because the most of them requires `Pass` and could not be tested separately from analyzer. 
> 2. Also I am afraid of "Helpers Anti-Pattern" 

> looks like compromise, something like
> - `checkers/helpers_len.go`
> - `checkers/helpers_bool.go`
> - `checkers/helpers_error.go`

> I can’t figure out what’s better and that’s why I left everything as it is.

> Therefore, I’m open to any proposals, just take into account the points above 👍

_Originally posted by @Antonboom in https://github.com/Antonboom/testifylint/pull/78#discussion_r1573176916_

ccoVeille:
I'm curious do you have a good article about "Helper Antipattern"

I found these via a quick google search:
- https://softwareengineering.stackexchange.com/questions/422907/is-putting-general-use-functions-in-a-helpers-file-an-anti-pattern-or-code-sme
- https://stackoverflow.com/questions/22049954/why-are-helperclasses-anti-pattern

Do you have any ?

I would like to talk about it on my https://dev.to/ccoveille once I will have understand it. I will group them with article about code pattern like the one that can be found here https://github.com/tmrts/go-patterns

PR #92, opened 2024-05-18
introduce helpers_*.go files

Antonboom:
Closes https://github.com/Antonboom/testifylint/issues/84


ccoVeille:
It would be way easier to understand Thanks

Issue #93, opened 2024-05-26
compares: shall suggest (Not)Same when comparing pointers


mmorel-35:
When two pointers are compared and expected to be different NotSame shall be suggested instead of NotEqual
I have two exemple for this : 

* https://github.com/jaegertracing/jaeger/pull/5134/files#diff-547fc06cb4531ef4316f167123091774e3dd7fb6dcc77a9b4f39f81b584cc43a
* https://github.com/vmware-tanzu/velero/pull/7671/files#diff-701da9642e59239de4f6bc012f4a307887935cd4414f6b9081d3b0a5bcf3b79d

ccoVeille:
Especially here

https://github.com/jaegertracing/jaeger/pull/5134#discussion_r1462749383

And here

https://github.com/vmware-tanzu/velero/pull/7671#discussion_r1615174861

ccoVeille:
I agree with your suggestion 👍

I commented PR for giving context

Antonboom:
@mmorel-35 @ccoVeille 

please, check patch with fix
```
go install github.com/Antonboom/testifylint@6b47034
```

on your code bases

ccoVeille:
I found nothing on my codebase and personal project that would raise "Same"

I have 1606 errors reported 282 files with errors over 1800+ test files

the number are unchanged between v1.3.0 and 6b47034

So I would say it's good

ccoVeille:
And I can also confirm that I'm able to spot the difference when launching the 2 versions on [jager code before fix](https://github.com/jaegertracing/jaeger/commit/ef38eabdb3b1bf0748ab9af93f1912830bd228b7) 

Before

```console
$ go vet -vettool="$(which testifylint)" ./... 2>&1 | fgrep 336
plugin/sampling/strategystore/static/strategy_store_test.go:336:2: compares: use assert.NotEqual
```

After
```console
$ go vet -vettool="$(which testifylint)" ./... 2>&1 | fgrep 336
plugin/sampling/strategystore/static/strategy_store_test.go:336:2: compares: use assert.NotSame
```

Issue #75, opened 2024-04-07
zero: use zero instead of `asserr.Equal(t, 0, v)`


ccoVeille:
For records, I started working on the implementation of this feature request.

https://github.com/Antonboom/testifylint/blob/master/CONTRIBUTING.md#zero

ccoVeille:
I've read the consideration about the doubt about using Zero vs Empty

> I'm not sure if anyone uses assert.Zero – it looks strange and conflicts with assert.Empty:
> Maybe it's better to make configurable support for other types in the empty checker and vice versa to prohibit the Zero?
> Source: https://github.com/Antonboom/testifylint/blob/master/CONTRIBUTING.md#zero

I'll explain in this issue what I plan to implement, so we could talk about the implementation here

ccoVeille:
I created the issue to keep a track of it, and allow other people to discuss it.

Stay tuned, and ping me if you see no progress in the next month.

ccoVeille:
For this reason, and because I struggle in making it works with `Empty` checker without conflicts in my current implementation, it seems simpler to add the Zero detection to existing `Empty` checker. What do you think about it ?

Antonboom:
@ccoVeille hi

I reread carefully the messages and examples above and looks like I was wrong

1. `Empty` is concentrated on equals with `len`.
2. `Zero` is concentrated on equals with zero value objects.

And these are separate checkers, and `zero` should cover not only zeros, but the full list:

<details>

```go
	zeros = []interface{}{
		false,
		byte(0),
		complex64(0),
		complex128(0),
		float32(0),
		float64(0),
		int(0),
		int8(0),
		int16(0),
		int32(0),
		int64(0),
		rune(0),
		uint(0),
		uint8(0),
		uint16(0),
		uint32(0),
		uint64(0),
		uintptr(0),
		"",
		[0]interface{}{},
		[]interface{}(nil),
		struct{ x int }{},
		(*interface{})(nil),
		(func())(nil),
		nil,
		interface{}(nil),
		map[interface{}]interface{}(nil),
		(chan interface{})(nil),
		(<-chan interface{})(nil),
		(chan<- interface{})(nil),
	}
```

</details>

If user uses `assert.Equal(t, 0, len(warningMsg))` then we suggest `assert.Empy(t, errors)`.
If uses uses `assert.Equal(t, "", warningMsg)` then we suggest `assert.Zero(t, warningMsg)`.

ccoVeille:
[...]

PR #96, opened 2024-05-27
compares: support Same/NotSame

Antonboom:
Closes https://github.com/Antonboom/testifylint/issues/93

### DIFF
--- a/internal/checkers/compares.go
+++ b/internal/checkers/compares.go
@@ -3,6 +3,7 @@ package checkers
 import (
 	"go/ast"
 	"go/token"
+	"go/types"
 
 	"golang.org/x/tools/go/analysis"
 )
@@ -26,6 +27,9 @@ import (
 //	assert.GreaterOrEqual(t, a, b)
 //	assert.Less(t, a, b)
 //	assert.LessOrEqual(t, a, b)
+//
+// If `a` and `b` are pointers then `assert.Same`/`NotSame` is required instead,
+// due to the inappropriate recursive nature of `assert.Equal` (based on `reflect.DeepEqual`).
 type Compares struct{}
 
 // NewCompares constructs Compares checker.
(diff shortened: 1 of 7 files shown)


ccoVeille:
I'll check later today, but the code looks good.

internal/checkers/compares.go
```diff
@@ -42,13 +46,24 @@ func (checker Compares) Check(pass *analysis.Pass, call *CallMeta) *analysis.Dia
...
-		tokenToProposedFn = tokenToProposedFnInsteadOfTrue
+		if ptrComparison {
+			tokenToProposedFn = tokenToProposedFnInsteadOfTrueForPtr
+		} else {
+			tokenToProposedFn = tokenToProposedFnInsteadOfTrue
+		}
+
 	case "False":
-		tokenToProposedFn = tokenToProposedFnInsteadOfFalse
+		if ptrComparison {
```
ccoVeille, line 62: This is the kind of implementation I was thinking for promoting zero over empty

https://github.com/Antonboom/testifylint/issues/75#issuecomment-2041380821

https://github.com/Antonboom/testifylint/issues/75#issuecomment-2067622523

internal/checkers/compares.go
```diff
@@ -42,13 +46,24 @@ func (checker Compares) Check(pass *analysis.Pass, call *CallMeta) *analysis.Dia
 		return nil
 	}
 
+	_, ptrComparison := pass.TypesInfo.TypeOf(be.X).(*types.Pointer)
+
 	var tokenToProposedFn map[token.Token]string
 
 	switch call.Fn.NameFTrimmed {
```
ccoVeille, line 53: This code works, but then it becomes very difficult to read. I mean a switch with 2 if in it.

Maybe code could be refactored to use a function to return the token suggestion.

Or the map, could have extra level.

Just a suggestion, you can of course discard

ccoVeille (approved):

ccoVeille:
I confirm it's OK, except tge few remarks I raised but that could be ignored, or refactored later

internal/checkers/compares.go
Antonboom: I have a "zero thread" in to-do :sorry:
complex story 😢

internal/checkers/compares.go
Antonboom: > Or the map, could have extra level.

I thought about it, but did not find the appropriate extendable key

But I simplified the code, thanks 👌

internal/checkers/compares.go
ccoVeille: OK, understood. I'm not very available either, haha

internal/checkers/compares.go
ccoVeille: I confirm, it's way better now. The code is good and clear. I don't think the map thing would have been as simple as your implementation. So I'm very happy with it.

Issue #95, opened 2024-05-27
error-nil: mixed struct "type" can lead to false-positive (v1.3.0)


ldez:
```console
$ golangci-lint run -E testifylint
main_test.go:32:2: error-nil: use assert.NoError (testifylint)
        assert.Nil(t, resp)
        ^
```

The example is a fake API client.

```go
package main

import (
	"testing"

	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
)

// APIResponse can be an error or not.
type APIResponse struct {
	Status   int
	Data     string
	ErrorMsg string
}

func (a APIResponse) Error() string {
	return a.ErrorMsg
}

func Update(a string) (*APIResponse, error) {
	if a == "a" {
		return &APIResponse{Status: 200, Data: "fake"}, nil
	}

	return nil, &APIResponse{Status: 500, ErrorMsg: "Oops"}
}

func TestName(t *testing.T) {
	resp, err := Update("b")
	require.Error(t, err, new(*APIResponse))
	assert.Nil(t, resp)
}
```

PR #97, opened 2024-05-27
helpers: isError: strict type comparison instead of implementing check

Antonboom:
Closes https://github.com/Antonboom/testifylint/issues/95


ccoVeille:
I'll try to perform the same analysis I made here

https://github.com/Antonboom/testifylint/issues/93#issuecomment-2133226047

And I will tell you if something change

ccoVeille:
> I'll try to perform the same analysis I made here
> 
> [#93 (comment)](https://github.com/Antonboom/testifylint/issues/93#issuecomment-2133226047)
> 
> And I will tell you if something change

Nothing changes on the code I have (number of error reported, and error reported), so I would say it's OK

PR #103, opened 2024-06-05
contribution: remove zero checker

Antonboom:
Closes https://github.com/Antonboom/testifylint/issues/75


ccoVeille:
Replying here as other is locked.

https://github.com/Antonboom/testifylint/issues/75#issuecomment-2149006927

OK, it's true that the existence of Zero vs Empty in testify is already a problem.

Suggesting to replace one my another would be a problem on code already using testifylint that would have to change valid tests.

So I'm OK with your choice

PR #107, opened 2024-06-06
useless-assert: add Zero and Empty to detected pattern

### LINKED ISSUE #105: useless-assert: add Empty And Zero
Some people may use testify Zero, Empty, NotZero, and NotEmpty asserters the same way True or False could have been used

ccoVeille:
Fixes #105


Antonboom:
I've never seen these somebody use bool operations in these assertions 

does this really make sense?

ccoVeille:
I would have said the same for EqualValues

I would say it will catch almost nothing, but why not merging it. It costs nothing to maintain.

PR #109, opened 2024-06-06
empty: added empty and zero length detection

### LINKED ISSUE #108: empty: add Zero(len) and Empty(len)
Surprisingly, this pattern is used very often

https://github.com/search?q=language%3Ago+%22assert.Zero%28t%2C+len%22&type=code

```go
			assert.Zero(t, len(result))
```

https://github.com/search?q=language%3Ago+%22assert.Empty%28t%2C+len%22&type=code

```go
			assert.Empty(t, len(result))
```

We should suggest using

```go
			assert.Empty(t, result)
```

ccoVeille:
Fixes #108


Antonboom (approved):
@ccoVeille, nice point, thanks!

And I am happy that checkers are easy extendable and you reused existing helper

Issue #110, opened 2024-06-06
 compares: add Empty And Zero


ccoVeille:
Some people may use testify Zero, Empty, NotZero, and NotEmpty asserters the same way True or False could have been used

Antonboom:
Hi!

Any real examples please, looks like something weird.
I do not want to promote the very strange use of `testify`, even in tests.

ccoVeille:
I failed to find real life example…

Should I nuke #111 completly ? or could we keep the refactoring commits I made ?

Antonboom:
> or could we keep the refactoring commits I made ?

Yes, please, open new PR with refactored part if this change makes sense
Thanks

PR #116, opened 2024-06-07
bool compare: add Zero and Empty

ccoVeille:
2 blocks of commits:

- Refactoring to prepare the next block
   - **gen: move assertion sorter to a function so it could be reused**
   - **bool_compare: use variables in generator**
   - **bool_compare: use predicting order for tests**
   - **bool_compare: refactor tests to easily add use cases**

- Add zero and empty detection to bool-compare

Issue #106, opened 2024-06-06
bool-compare: compares: add Empty And Zero


ccoVeille:
Some people may use testify Zero, Empty, NotZero, and NotEmpty asserters the same way True or False could have been used

ccoVeille:
I thought about it after we closed https://github.com/Antonboom/testifylint/issues/75

Zero and Empty are like True and False, they can be used wrongly

This one is somehow related to #105 as it's about adding zero and empty to existing asserter.

I preferred a dedicated issue as they could be reviewed separately

ccoVeille:
I have a PR almost ready for this one I think, I will finish #105 first

Antonboom:
@ccoVeille, I did not catch what patterns are you talking about?

Can you share specific examples and links to search, thanks

ccoVeille:
My idea was to extend existing checker with strange case people may use as empty/zero could be used the same way that True/False are used in bool compare.

I checked at existing piece of code by using GitHub search and I found nothing relevant.

You seem not likely open to catch this. OK then.

Issue #94, opened 2024-05-26
negative-positive: support typed `0`


ccoVeille:
another yet [case](https://github.com/jaegertracing/jaeger/pull/5134/files#diff-83ecac470d767b6e86892d42d2a07581549b6280e355fb316c2e30332bd9d9d1R123) of `assert.Positive`:
```
assert.Greater(t, gs[lastValueLogCleanedName], int64(0))
```

_Originally posted by @Antonboom in https://github.com/Antonboom/testifylint/issues/93#issuecomment-2132368054_

The idea is to complete the linter added with #76

ccoVeille:
I have found this one also in the quoted PR


```go
        assert.Greater(t, stopwatch.ElapsedTime(), time.Duration(0))
```

I'm curious how testifylint behaves right now

ccoVeille:
Just to make sure you wanted to talk about typed 0 ?

I'm unsure what the typed nil could refer to

ccoVeille:
![8tbabj.jpg](https://github.com/Antonboom/testifylint/assets/3875889/f874084d-ba2f-489a-96b5-f4ba256fda20)

Antonboom:
My bad, title is fixed 👌

ccoVeille:
@Antonboom I'm looking at the code and I see similarities between

isIntNumber and isIntBasicLit

https://github.com/Antonboom/testifylint/blob/master/internal/checkers/helpers_basic_type.go#L18-L31

I don't get the difference of comparing the number to its Sprintf value.

The only use case, I can think about is -0 vs 0, but then I don't understand

I looked when the `Sprintf("%d", v) == v` was introduced and I found d6f0ef8f

But I see no reason to add it in the context of this commit, do you have some context about it, because otherwise I might refactor a bit isIntBasicLit

PR #121, opened 2024-06-09
refactor generator

ccoVeille:
- **gen: move assertion sorter to a function so it could be reused**
- **bool_compare: refactor tests**

These commits were part of ~#107~ and #111 that were suggested changes were discarded, but the refactoring made in each PR could be kept as discussed [here](https://github.com/Antonboom/testifylint/issues/110#issuecomment-2155469152)


ccoVeille:
This PR uses commits available in:
- #116 created for #106
- #111 created for #110

ccoVeille:
@Antonboom do you think we could merge this one, it would help me to refactor for 
- #94

internal/testgen/main.go
```diff
@@ -135,3 +136,13 @@ func genGoFileFromTmpl(output string, exec Executor, data any) error {
 
 	return os.WriteFile(output, formatted, 0o644)
 }
+
+func sortAssertions(assertions []Assertion) {
```
Antonboom, line 140: move it to `assertion.go`

let's keep `main.go` free from any helpers, the file is about generator itself

internal/testgen/gen_bool_compare.go
```diff
@@ -26,6 +26,140 @@ func (g BoolCompareTestsGenerator) TemplateData() any {
 		ValidAssertions   []Assertion
 	}
 
+	ignoredAssertions := make([]Assertion, 0)
```
Antonboom, line 29: https://github.com/uber-go/guide/blob/master/style.md#nil-is-a-valid-slice

or write
```go
var ignoredAssertions []Assertion
```

or use real capacity
```go
ignoredAssertions := make([]Assertion, 2*8+6)
// but! magic numbers + no make sense for this case anyway
```

internal/testgen/gen_bool_compare.go
```diff
@@ -26,6 +26,140 @@ func (g BoolCompareTestsGenerator) TemplateData() any {
...
+	ignoredAssertions = append(ignoredAssertions,
+		Assertion{Fn: "Equal", Argsf: "foo, foo"},
+		Assertion{Fn: "NotEqual", Argsf: "foo, foo"},
+		Assertion{Fn: "True", Argsf: "foo == foo"},
+		Assertion{Fn: "False", Argsf: "foo == foo"},
+		Assertion{Fn: "True", Argsf: "foo != foo"},
+		Assertion{Fn: "False", Argsf: "foo != foo"},
+	)
+
+	invalidAssertionsForTrue := make([]Assertion, 0)
```
Antonboom, line 87: var

internal/testgen/gen_bool_compare.go
```diff
@@ -26,6 +26,140 @@ func (g BoolCompareTestsGenerator) TemplateData() any {
...
+	invalidAssertionsForFalse := make([]Assertion, 0)
```
Antonboom, line 88: var

internal/testgen/gen_compares.go
```diff
@@ -22,41 +22,52 @@ func (g ComparesTestsGenerator) TemplateData() any {
...
+	}
+
+	invalidAssertions = append(invalidAssertions,
+		Assertion{Fn: "True", Argsf: "ptrA == ptrB", ReportMsgf: report, ProposedFn: "Same", ProposedArgsf: "ptrA, ptrB"},
+		Assertion{Fn: "True", Argsf: "ptrA != ptrB", ReportMsgf: report, ProposedFn: "NotSame", ProposedArgsf: "ptrA, ptrB"},
+		Assertion{Fn: "False", Argsf: "ptrA == ptrB", ReportMsgf: report, ProposedFn: "NotSame", ProposedArgsf: "ptrA, ptrB"},
+		Assertion{Fn: "False", Argsf: "ptrA != ptrB", ReportMsgf: report, ProposedFn: "Same", ProposedArgsf: "ptrA, ptrB"},
+	)
+
+	sortAssertions(invalidAssertions)
```
Antonboom, line 60: any reason to sort it?
the order of assertions is determined

internal/testgen/gen_useless_assert.go
```diff
@@ -71,13 +70,7 @@ func (g UselessAssertTestsGenerator) TemplateData() any {
...
 	}
 
-	sort.Slice(twoSideAssertions, func(i, j int) bool {
-		lhs, rhs := twoSideAssertions[i], twoSideAssertions[j]
-		if lhs.Fn == rhs.Fn {
-			return lhs.Argsf < rhs.Argsf
-		}
-		return lhs.Fn < rhs.Fn
-	})
+	sortAssertions(twoSideAssertions)
```
Antonboom, line 73: let's look we can just leave the sort code here only

because there is `map` above

internal/testgen/gen_bool_compare.go
```diff
@@ -26,6 +26,140 @@ func (g BoolCompareTestsGenerator) TemplateData() any {
...
+		invalidAssertionsForTrue = append(invalidAssertionsForTrue,
+			Assertion{Fn: "True", Argsf: argsf, ReportMsgf: reportSimplify, ProposedArgsf: "predicate"},
+		)
+
+		invalidAssertionsForFalse = append(invalidAssertionsForFalse,
+			Assertion{Fn: "False", Argsf: argsf, ReportMsgf: reportSimplify, ProposedArgsf: "predicate"},
+		)
+	}
+
+	sortAssertions(invalidAssertionsForTrue)
```
Antonboom, line 159: any reason to sort it?
the order of assertions is determined

safe this PR from unrelated changes around generated test files

internal/testgen/gen_compares.go
```diff
@@ -22,41 +22,52 @@ func (g ComparesTestsGenerator) TemplateData() any {
...
-		ignored = append(ignored,
+		ignoredAssertions = append(ignoredAssertions,
 			Assertion{Fn: "True", Argsf: fmt.Sprintf("c %s d", tok)},
 			Assertion{Fn: "False", Argsf: fmt.Sprintf("d %s c", tok)},
 		)
 	}
 
+	var invalidAssertions []Assertion //nolint:prealloc
+	for checker, oppositeChecker := range map[string]string{
+		"True": "False",
```
Antonboom, line 35: loop and map doesn't make sense in such case
this is only increase complexity of support

please revert it or do in the same way as `append` below

Antonboom (changes requested):

internal/testgen/main.go
ccoVeille: Sure. Thanks for the review

ccoVeille:
I pushed too fast sorry.

I will fix other too

internal/testgen/gen_compares.go
ccoVeille: This one was there, because #111 had another commit to add things 1eb5934a8f43d09422d826ac93987ae2e40e882d

I will revert sure

internal/testgen/gen_useless_assert.go
ccoVeille: OK, I will remove the others

internal/testgen/gen_bool_compare.go
ccoVeille: I will remove it then

ccoVeille:
I resolved all your comments @Antonboom please check and review.

The number of changes is then very limited

PR #124, opened 2024-06-14
Suggest NotEmpty for Positive len

### LINKED ISSUE #122: len+compare: false-positive on `assert.True(t, len(myarray) > 0)`
With `assert.True(t, len(myarray) > 0)`, it is replaced by `assert.Positive(t, len(myarray))` with compare rule
I would then expect len or empty rule to suggest 
`assert.NotEmpty(t, myarray)`

Same logic with `assert.True(t, len(myarray) >= 1)`  is replaced by `assert.GreaterOrEqual(t, len(myarray), 1)`, shall then be `assert.NotEmpty(t, myarray)`

mmorel-35:
Closes #122


ccoVeille (approved):
LGTM

PR #126, opened 2024-06-17
negative-positive: support typed 0

### LINKED ISSUE #94: negative-positive: support typed `0`

ccoVeille:
Fixes #94


ccoVeille:
I could iterate and make the commit atomic, so it won't be a block like this.

I created a isZeroValue separated from isZero to avoid conflicting with code that is already using isZero or isIntNumber

Tell me what you think about it @Antonboom

README.md
```diff
@@ -565,14 +565,53 @@ assert.Len(t, arr, 3)
 ```go
 ❌
 assert.Less(t, a, 0)
+assert.Less(t, a, int(0))
```
Antonboom, line 568: I appreciated that you follows guidelines, but I propose to do exception here.

just add to description 

```
**Reason**: More appropriate `testify` API with clearer failure message.

Typed zeros (like `int8(0)`) are also supported.

---

```

and remove such detailed examples from README and GoDoc, to make it more readable

README.md
```diff
@@ -565,14 +565,53 @@ assert.Len(t, arr, 3)
...
+assert.Greater(t, int8(0), a)
+assert.Greater(t, int16(0), a)
+assert.Greater(t, int32(0), a)
+assert.Greater(t, int64(0), a)
+assert.Greater(t, uint(0), a)
+assert.Greater(t, uint8(0), a)
+assert.Greater(t, uint16(0), a)
+assert.Greater(t, uint32(0), a)
+assert.Greater(t, uint64(0), a)
 assert.True(t, a < 0)
```
Antonboom, line 589: why no changes around expression in True/False?

this is relevant for this too, isn't it?

Antonboom (changes requested):

README.md
ccoVeille: I think I suggested it in another package and you would have said, we had no examples of such usage.

I might have misunderstood also.

But I will add then then, sure

README.md
ccoVeille: Sure, I will

ccoVeille:
I made the changes. Let me know if you are fine with them

README.md
```diff
@@ -587,6 +587,8 @@ assert.Positive(t, a)
 **Enabled by default**: true <br>
 **Reason**: More appropriate `testify` API with clearer failure message.
 
+Typed zeros (like `int(0)`, `int(8)`, … `uint(64)`) are also supported.
```
Antonboom: typo – must be `(0)`, not `(8), (64)`

internal/checkers/negative_postive.go
```diff
@@ -25,6 +25,8 @@ import (
 //	assert.False(t, a <= 0)
 //	assert.False(t, 0 >= a)
 //
+// Typed zeros (like `int(0)`, `int(8)`, … `uint(64)`) are also supported.
```
Antonboom: invalid place of comment – you broke the sentence "detects ... and requires ..."

internal/testgen/gen_negative_positive.go
```diff
@@ -19,28 +20,58 @@ func (g NegativePositiveTestsGenerator) TemplateData() any {
...
+			Assertion{Fn: "Less", Argsf: fmt.Sprintf("a, %s", zero), ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+			Assertion{Fn: "Greater", Argsf: fmt.Sprintf("%s, a", zero), ReportMsgf: report, ProposedFn: "Negative", ProposedArgsf: "a"},
+
+			Assertion{Fn: "Greater", Argsf: fmt.Sprintf("a, %s", zero), ReportMsgf: report, ProposedFn: "Positive", ProposedArgsf: "a"},
+			Assertion{Fn: "Less", Argsf: fmt.Sprintf("%s, a", zero), ReportMsgf: report, ProposedFn: "Positive", ProposedArgsf: "a"},
+		)
+	}
+
+	for zero, a := range map[string]string{
+		"0":         "a",
```
Antonboom: can be done in one loop for slices of types

internal/testgen/gen_negative_positive.go
```diff
@@ -19,28 +20,58 @@ func (g NegativePositiveTestsGenerator) TemplateData() any {
 		report  = checker + ": use %s.%s"
 	)
 
+	var invalidAssertions []Assertion
+
+	for _, zero := range []string{
+		"0",
+		"int(0)", "int8(0)", "int16(0)", "int32(0)", "int64(0)",
+		"uint(0)", "uint8(0)", "uint16(0)", "uint32(0)", "uint64(0)",
```
Antonboom: Negative/Positive looks useless for unsigned values.

they are always Positive and never Negative

looks like a case for `incorrect-assert` checker (not implemented yet)

internal/testgen/gen_negative_positive.go
```diff
@@ -19,28 +20,58 @@ func (g NegativePositiveTestsGenerator) TemplateData() any {
...
+		"int(0)":    "a",
+		"int8(0)":   "int8A",
+		"int16(0)":  "int16A",
+		"int32(0)":  "int32A",
+		"int64(0)":  "int64A",
+		"uint(0)":   "uintA",
+		"uint8(0)":  "uint8A",
+		"uint16(0)": "uint16A",
+		"uint32(0)": "uint32A",
+		"uint64(0)": "uint64A",
```
Antonboom: not covered edge cases 
```
uint(value), 0
CustomInt(value), 0
```

Antonboom (commented):
@ccoVeille please, be more attentive

and review PR again after my fixes 👌

ccoVeille:
Thanks for the  test refactoring.

Also the changes to generate and test one checker without the others 🥰

I don't get the point of removing the unsigned int? I'm OK with it, but I'm unsure why.

Antonboom:
@ccoVeille read the comments above 👆

README.md
ccoVeille: Indeed 🤦

internal/testgen/gen_negative_positive.go
ccoVeille: Indeed, it's obvious when you spot it.

Unsigned int cannot be negative.

But, unsigned int can be positive, so they leads to compare they are not zero/empty.

To be defined, because then it becomes closed to zero (more than empty)

ccoVeille:
I'm OK with the changes, you removed the uint vs negative, which is obvious

For records, I found examples of code where uint is used in something that should be converted to `Positive`

https://github.com/search?q=language%3Ago+%22assert.Less%28t%2C+uint%22&type=code

https://github.com/mysteriumnetwork/node/blob/adb67589d42929b8ad6f3a76177d585e4f96b7e4/consumer/entertainment/estimator_test.go#L45
```go
assert.Less(t, uint64(0), e.VideoMinutes)
```

https://github.com/mehrdadrad/tcpprobe/blob/782fdd0b0962d91495e3ec02341f8d5c8915ea2e/tp_test.go#L61-L62
```go
	assert.Less(t, uint32(0), c.stats.Rto)
	assert.Less(t, uint32(0), c.stats.Ato)
```

https://github.com/yl2chen/cidranger/blob/d1cb2c52f37ac257db665ddfaf4af22e89863471/trie_test.go#L545C1-L545C41
```go
	assert.Less(t, uint64(0), baseLineHeap)
```

These are edge cases, but they are legitimate according to me.

But we could move them to a separate issue, and current PR could be merged like this.

Taskfile.yml
```diff
@@ -46,6 +46,12 @@ tasks:
       - echo "Test..."
       - go test ./...
 
+  # task test:checker -- negative-positive
+  test:checker:
+    deps: [ test:gen ]
```
ccoVeille, line 51: thanks for adding the generation here too, I tried yesterday and I faced misunderstanding :+1:

ccoVeille:
> @ccoVeille
> 
> > But, unsigned int can be positive, so they leads to compare they are not zero/empty.
> 
> this is just my mistake. I returned back `uint` support for `assert.Positive` and added real-life examples to tests

I think we are good

Issue #82, opened 2024-04-17
regexp: simplify `assert.Regexp(t, regexp.MustCompile)`


ccoVeille:
@dolmen post helped me to spot a new checker that could be added

https://github.com/Antonboom/testifylint/issues/81#issuecomment-2060618745

assert.Regexp(t, regex.MustCompile("whatever")… could be simplified into assert.Regexp(t, "whatever"…

I spotted it here
https://github.com/SCP002/m3u_merge_astra/blob/9ae08f40f415292bf646a2dd7811f4d5d799dea6/src/util/logger/logger_test.go#L23

ccoVeille:
Also cc @kevinburkesegment who somehow initiated my discovery

Related: stretchr/testify#1585

ccoVeille:
> Grafana has a fork from regexp, see https://github.com/grafana/regexp
> 
> It might be intrested to suggest it for the standard library and grafana’s version as well ?

I'm unsure to get your point.

There are some interesting lib for regexp.

But, here we are in a linter specialized in: tests (so perf is not the priority) and testify.

Testify has an asserter for Regexp.
This issue is about converting

```go
assert.Regexp(t, regex.MustCompile("whatever"))

// simplified

assert.Regexp(t, "whatever"…
```

kevinburkesegment:
I do not think that adding a new dependency would simplify things

ccoVeille:
Now, I get it.

I didn't check the implementation, but I'm doubtful `assert.Regexp` works with something else than std lib.

But then, why not, did you find examples of such usage?

Antonboom:
Please, stop this discussion.
`testifylint` should work with std lib only or libs that inherited from testify's assertions signatures.

PR #136, opened 2024-06-20
new checker 'regexp'

Antonboom:
Closes https://github.com/Antonboom/testifylint/issues/82

But I wonder that we lost syntax highlighting 🤔 

Screenshot 2024-06-20 at 20 35 48

P.S. https://youtrack.jetbrains.com/issue/GO-17040/Proposal-highlight-regexp-string-in-testifys-assert.Regexp-NotRegexp


internal/checkers/regexp.go
```diff
@@ -0,0 +1,42 @@
...
+//	assert.Regexp(t, `\[.*\] DEBUG \(.*TestNew.*\): message`, out)
+//	assert.NotRegexp(t, `\[.*\] TRACE message`, out)
+type Regexp struct{}
+
+// NewRegexp constructs Regexp checker.
+func NewRegexp() Regexp     { return Regexp{} }
+func (Regexp) Name() string { return "regexp" }
+
+func (checker Regexp) Check(pass *analysis.Pass, call *CallMeta) *analysis.Diagnostic {
+	if n := call.Fn.NameFTrimmed; n != "Regexp" && n != "NotRegexp" {
```
ccoVeille, line 25: When checking the same variable multiple times, I would have used a switch here.
I think it would improve readability, especially because these are negation.

ccoVeille (commented):
Just one readability remark, but LGTM 👍

Btw `isPkgFnCall` is a great addition 🎉

Issue #138, opened 2024-06-20
error-nil: add Empty and Zero


ccoVeille:
we do not detect error used with empty/zero, right now

```go
var err error
assert.Empty(t, err)
assert.Zero(t, err)
```

We should recommend using `NoError`

```go
var err error
assert.NoError(t, err)
```

this pattern is widely used https://github.com/search?q=language%3Ago+%22assert.Empty%28t%2C+err%22&type=code
https://github.com/search?q=language%3Ago+%22assert.Zero%28t%2C+err%29%22&type=code

These two are candidates for `error-nil`

_Originally posted by @ccoVeille in https://github.com/Antonboom/testifylint/issues/119#issuecomment-2181588845_

Antonboom:
Yes, this is valid for me.

Need to add to ignored cases in tests:
```go
assert.Empty(t, errs)
assert.Empty(t, err.String())
```

PR #153, opened 2024-06-24
error-nil: support Empty and Zero error

mmorel-35:
This is a partial fix for #138

It doesn't fix this situations

```go
assert.Empty(t, errs)
assert.Empty(t, err.String())
```


internal/testgen/gen_error_nil.go
```diff
@@ -76,6 +80,10 @@ func (g ErrorNilTestsGenerator) TemplateData() any {
 			{Fn: "Equal", Argsf: "nil, nil"},
 			{Fn: "NotEqual", Argsf: "err, err"},
 			{Fn: "NotEqual", Argsf: "nil, nil"},
+			{Fn: "Empty", Argsf: "err.Error()"},    // Requested by https://github.com/Antonboom/testifylint/issues/138
+			{Fn: "NotEmpty", Argsf: "err.Error()"}, // Requested by https://github.com/Antonboom/testifylint/issues/138
+			{Fn: "Zero", Argsf: "err.Error()"},     // Requested by https://github.com/Antonboom/testifylint/issues/138
+			{Fn: "NotZero", Argsf: "err.Error()"},  // Requested by https://github.com/Antonboom/testifylint/issues/138
```
ccoVeille, line 86: These one are not the one requested by Anton I think

He asked for err.String so the [stringer](https://pkg.go.dev/fmt#Stringer) interface, not the [Error](https://go.dev/ref/spec#Errors) one.

These one are valid Error that the error-nil could/should catch, to be discussed of course.

ccoVeille (commented):
I'm unsure if it's normal you used err.String in Ignored.

We need to [discuss with Anton about it](https://github.com/Antonboom/testifylint/pull/153/files#r1650927950)

internal/testgen/gen_error_nil.go
ccoVeille: @Antonboom what did you mean with "err.String()" and what should be done with `Empty(err.Error())`

ccoVeille: The pattern is used https://github.com/search?q=language%3Ago+%22assert.Empty%28t%2C+err.Error%28%29%29%22&type=code

I found nothing that detects it

ccoVeille: anyway @mmorel-35  , I'm fine if these are not added, first and marked as to-do later, as you said it was a partial support

Antonboom: > what did you mean with "err.String()"

It just was a typo, thanks.

> should error-nil catch it ?

I think doesn't, because this is completely difference assertions. Nil error != empty message of not-nil error.

PR #155, opened 2024-06-25
new checker `http-const`

### LINKED ISSUE #141: http-const: suggest `http` package constants in HTTP asserts
```go
❌   assert.HTTPStatusCode(t, handler, "GET", "/index", nil, 200)
     assert.HTTPBodyContains(t, handler, "GET", "/index", nil, "counter")
     // etc.

✅   assert.HTTPStatusCode(t, handler, http.MethodGet, "/index", nil, http.StatusOK)
     assert.HTTPBodyContains(t, handler, http.MethodGet, "/index", nil, "counter")
```

Is similar to the [usestdlibvars](https://golangci-lint.run/usage/linters/#usestdlibvars) linter.

mmorel-35:
Closes #141


internal/checkers/helpers_basic_type.go
```diff
@@ -56,9 +56,14 @@ func isTypedIntNumber(e ast.Expr, v int, types ...string) bool {
 	return false
 }
 
-func isIntNumber(e ast.Expr, v int) bool {
+func safeTypedBasicLit(e ast.Expr, typ token.Token) (*ast.BasicLit, bool) {
```
ccoVeille, line 59: I'm unsure about the name, but thanks I had also noticed this could be refactored into a dedicated function

internal/checkers/helpers_http.go
```diff
@@ -3,12 +3,95 @@ package checkers
...
+	http.MethodPut:     "http.MethodPut",
+	http.MethodPatch:   "http.MethodPatch",
+	http.MethodDelete:  "http.MethodDelete",
+	http.MethodConnect: "http.MethodConnect",
+	http.MethodOptions: "http.MethodOptions",
+	http.MethodTrace:   "http.MethodTrace",
+}
+
+var httpStatusCode = map[string]string{
+	strconv.Itoa(http.StatusContinue):           "http.StatusContinue",
```
ccoVeille, line 27: I don't think we should consider performance for this

I would have used a `httpStatusCode = map[int]string{`

so no, strconv.Itoa here

but only when needed in the code, you can also create a dedicated testing helper if your prefer

but current code looks strange to me

ccoVeille (approved):
Thanks a lot, it would be great to have this linter

I was thinking about coding it, but you did it.

It's great!!!

internal/checkers/helpers_basic_type.go
ccoVeille: I'm so lame at wording. I'm only able to argue when I find a wording, but I'm never able to find a descent one to suggest 😅

internal/checkers/helpers_http.go
ccoVeille: Maybe Anton will provide his guidance

internal/checkers/http_const.go
```diff
@@ -0,0 +1,112 @@
...
+							Pos:     bt.Pos(),
+							End:     bt.End(),
+							NewText: []byte(newVal),
+						},
+					},
+				})
+			}
+		}
+		if len(suggestedFixes) > 0 {
+			return &analysis.Diagnostic{
```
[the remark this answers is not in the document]
ccoVeille: I agree

internal/checkers/http_const.go
```diff
@@ -0,0 +1,122 @@
...
+func (checker HTTPConst) Check(_ *analysis.Pass, call *CallMeta) *analysis.Diagnostic {
+	switch call.Fn.NameFTrimmed {
+	case "HTTPBody",
+		"HTTPBodyContains",
+		"HTTPBodyContainsf",
+		"HTTPBodyNotContains",
+		"HTTPBodyNotContainsf",
+		"HTTPError",
+		"HTTPErrorf",
+		"HTTPRedirect",
```
ccoVeille, line 36: No HTTPRedirectf ?

https://pkg.go.dev/github.com/stretchr/testify/assert#HTTPRedirectf

ccoVeille: Wait, you are using trimmed variable, please remove all the one with a trailing f

internal/checkers/http_const.go
```diff
@@ -0,0 +1,122 @@
...
+		"HTTPSuccess",
+		"HTTPSuccessf":
+
+		if len(call.Args) < 2 {
+			return nil
+		}
+
+		if bt, ok := safeTypedBasicLit(call.Args[1], token.STRING); ok {
```
ccoVeille, line 44: I would move this to a dedicated function, sane for http code detection. I would be easier to read, you pass arg1 and that's it. You can use early return everywhere

internal/checkers/http_const.go
```diff
@@ -0,0 +1,122 @@
...
+			currentVal := unquoteBasicLitValue(bt)
+			key := strings.ToUpper(currentVal)
+			if newVal, ok := httpMethod[key]; ok {
+				suggestedFixes = append(suggestedFixes, newConstReplacement(bt, currentVal, newVal))
+			}
+		}
+		if bt, ok := safeTypedBasicLit(call.Args[4], token.INT); ok {
+			currentVal := bt.Value
+			key := strings.ToUpper(currentVal)
```
ccoVeille, line 66: I was looking for this, please add tests for for lower case "get" or Pascal case (Get) and all other methods

ccoVeille (approved):

ccoVeille (commented):
I forgot to send this. It was pending

ccoVeille:
I know understand why @ldez (hi :wave:) expects PR to do not rewriting history during code review.

It's something I have been doing for a while, even in this project. But now I'm facing the way @mmorel-35 does, which I would have done too, I can see how difficult it is yo track the changes made between each iterations.

Here there is only one commit, and I have to read everything again and again instead of seeing the small changes made by each reviews

I repeat it's something I have done many times, even recently, so I don't blame you @mmorel-35, but now I understand Ludovic's guidance

internal/checkers/http_const.go
```diff
@@ -0,0 +1,128 @@
...
+		"HTTPBodyContains",
+		"HTTPBodyNotContains",
+		"HTTPError",
+		"HTTPRedirect",
+		"HTTPSuccess":
+
+		if len(call.Args) < 2 {
+			return nil
+		}
+		if suggestedFix := newHTTPMethodReplacement(call); suggestedFix != nil {
```
ccoVeille, line 40:
```suggestion
suggestedFix := newHTTPMethodReplacement(call)
if suggestedFix == nil {
        return nil
}

suggestedFixes = append(suggestedFixes, *suggestedFix)

```

ccoVeille: Wait, are you sure suggestedFixes ia empty before returning?

ccoVeille: Maybe the fix for the extra and the potential bug is to remove all the else branch, including the return

Please check if there is no other return that could break the pending suggestedFixes

internal/checkers/http_const.go
```diff
@@ -0,0 +1,128 @@
...
+			suggestedFixes = append(suggestedFixes, *suggestedFix)
+		}
+		if suggestedFix := newHTTPStatusCodeReplacement(call); suggestedFix != nil {
+			suggestedFixes = append(suggestedFixes, *suggestedFix)
+		}
+	}
+	if len(suggestedFixes) > 0 {
```
ccoVeille, line 56:
```suggestion
if len(suggestedFixes) == 0 {
     return nil
}
return newHTTPConstDiagnostic(checker.Name(), call, suggestedFixes)
```

internal/checkers/http_const.go
```diff
@@ -0,0 +1,128 @@
...
+		return newHTTPConstDiagnostic(checker.Name(), call, suggestedFixes)
+	}
+	return nil
+}
+
+func newHTTPMethodReplacement(call *CallMeta) *analysis.SuggestedFix {
+	if bt, ok := safeTypedBasicLit(call.Args[1], token.STRING); ok {
```
ccoVeille, line 63: Please, stop doing function-wide if statement. It's harder to read and to maintain


```suggestion
bt, ok := safeTypedBasicLit(call.Args[1], token.STRING)

if !ok {
return nil
}
```

I recommend using revive that could detect this pattern. early-return is one we need I think

internal/checkers/http_const.go
```diff
@@ -0,0 +1,128 @@
...
+		key := strings.ToUpper(currentVal)
+		if newVal, ok := httpMethod[key]; ok {
+			return newConstReplacement(bt, currentVal, newVal)
+		}
+	}
+	return nil
+}
+
+func newHTTPStatusCodeReplacement(call *CallMeta) *analysis.SuggestedFix {
+	if bt, ok := safeTypedBasicLit(call.Args[4], token.INT); ok {
```
ccoVeille, line 74:
```suggestion
bt, ok := safeTypedBasicLit(call.Args[4], token.INT);
if !ok {
return nil
}
```

ccoVeille (changes requested):

ccoVeille:
@mmorel-35 I have no idea what you changed with your last commit. I reviewed everything and spotted things I found. But I have no idea if it's new code or not.

ccoVeille:
Thanks for taking my workflow suggestion into account.

I do hate to see a log like that. But for reviewing it's way simpler.

Please note, I'm OK with you to rearrange the commits in the way you intend. But many only once you get an approval for the PR. Anx with a message like in the PR like "squashing previously reviewd changes"

I dislike squashed PR too, that's why I'm suggesting that.

Also squashing when merging would nuke the respective contribution of each developer. Think about Anton "maintainer review" commit.

At least, GitHub doesn't nuke authors who contributed as it adds the `Co-Authored-By ` tag in the squashed commit message.

But I do think it's way easier for reviewing with atomic and iterative commits.

Thanks again for baring with me

README.md
```diff
@@ -566,6 +567,24 @@ P.S. Look at [testify's issue](https://github.com/stretchr/testify/issues/772),
...
+assert.HTTPBodyContains(t, handler, "GET", "/index", nil, "counter")
+
+✅
+assert.HTTPStatusCode(t, handler, http.MethodGet, "/index", nil, http.StatusOK)
+assert.HTTPBodyContains(t, handler, http.MethodGet, "/index", nil, "counter")
+```
+
+**Autofix**: true. <br>
+**Enabled by default**: true. <br>
+**Reason**: Is similar to the [usestdlibvars](https://golangci-lint.run/usage/linters/#usestdlibvars) linter. <br>
```
ccoVeille, line 584: This one is for @Antonboom, Matthieu copy-pasted, and so moved the section available in Contributions.md

While the Reason was OK for a Dev purpose, I would expect a few more information by being a random Gopher finding this line. people are reading fast, and laziness/business might lead to do not click on the link.

So maybe if you could think about a better idea

At least something like that it doesn't have to be more

```suggestion
**Reason**: Cleaner code, this ckecker is similar to the [usestdlibvars](https://golangci-lint.run/usage/linters/#usestdlibvars) linter. <br>
```

internal/checkers/helpers_http.go
ccoVeille: @Antonboom it's me the serial mention guy 😅😁

What are your thoughts about this?

The fact the code imported from usedstdvars is a bit strange

internal/testgen/gen_http_const.go
```diff
@@ -0,0 +1,109 @@
...
+				Argsf:         `httpHelloName, "GET",  "/", url.Values{"name": []string{"World"}}, "Hello, World!"`,
+				ReportMsgf:    checker + ": use net/http constants instead of value",
+				ProposedArgsf: `httpHelloName, http.MethodGet,  "/", url.Values{"name": []string{"World"}}, "Hello, World!"`,
+			},
+		},
+		ValidAssertions: []Assertion{
+			{Fn: "HTTPStatusCode", Argsf: `httpOK, http.MethodGet, "/index", nil, http.StatusOK`},
+			{Fn: "HTTPBodyContains", Argsf: `httpHelloName, http.MethodGet,  "/", url.Values{"name": []string{"World"}}, "Hello, World!"`},
+		},
+		IgnoredAssertions: []Assertion{},
```
ccoVeille, line 53: I would add to the ignored assertions an unofficial code and method to make sure code won't catch things it shouldn't

                                Argsf:         `httpOK, "FOO", "/index", nil, 123`,

ccoVeille (approved):

ccoVeille:
Sute, thanks for letting us know.

internal/checkers/http_const.go
```diff
@@ -0,0 +1,129 @@
...
+		Pos:            rng.Pos(),
+		End:            rng.End(),
+		Category:       checker,
+		Message:        checker + ": " + msg,
+		SuggestedFixes: suggestedFixes,
+	}
+}
+
+func unquoteBasicLitValue(basicLit *ast.BasicLit) string {
+	s, _ := strconv.Unquote(basicLit.Value)
```
ccoVeille, line 127: I'm a bit surprised by the fact the current implementation ignore the error.

https://cs.opensource.google/go/go/+/go1.22.4:src/strconv/quote.go;l=382

There are a large number of use cases where it can returns an `strconv.ErrSyntax`

I would expect to handle the error in the code calling this method

ccoVeille: I would do this

```go
func unquoteBasicLitValue(basicLit *ast.BasicLit) (string, error) {
     return strconv.Unquote(basicLit.Value)
}
```

and when calling the code I would that
```go
	currentVal, err := unquoteBasicLitValue(bt)
	if err != nil {
	    return nil
	}
```

But then the method is somehow useless.

or to respect the pattern of the `value, ok` we have in other methods, I would do that


```go
func unquoteBasicLitValue(basicLit *ast.BasicLit) (string, bool) {
     value, err := strconv.Unquote(basicLit.Value)
     if err != nil {
          return "", false
     }
     return value, true
}
```

and when calling the code I would that
```go
	currentVal, err := unquoteBasicLitValue(bt)
	if err != nil {
	    return nil
	}
```

ccoVeille (approved):
LGTM thanks

internal/testgen/gen_http_const.go
ccoVeille: @mmorel-35 I don't know if you saw this one

ccoVeille (changes requested):

internal/testgen/gen_http_const.go
ccoVeille: oh you are editing in a browser, I didn't get it.

I can take care of it

ccoVeille: I think you should add me to maintainer/users of your project. 

```console
$ git push --no-verify --force-with-lease mmorel 
ERROR: Permission to mmorel-35/testifylint.git denied to ccoVeille.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights and the repository exists.
```

Anton can do updates, but because he is a maintainer of Antonboom/testifylint, he didn't grant us to be maintainers of ` Antonboom/testifylint`.

ccoVeille: Same error message

ccoVeille: I had to click on the email I didn't notice.

It works now

ccoVeille (approved):
LGTM 👍


May I suggest you to squash/fix up now? To reduce the number of commits

README.md
```diff
@@ -566,6 +567,24 @@ P.S. Look at [testify's issue](https://github.com/stretchr/testify/issues/772),
...
+assert.HTTPBodyContains(t, handler, "GET", "/index", nil, "counter")
+
+✅
+assert.HTTPStatusCode(t, handler, http.MethodGet, "/index", nil, http.StatusOK)
+assert.HTTPBodyContains(t, handler, http.MethodGet, "/index", nil, "counter")
+```
+
+**Autofix**: true. <br>
+**Enabled by default**: true. <br>
+**Reason**: Cleaner code, this ckecker is similar to the [usestdlibvars](https://golangci-lint.run/usage/linters/#usestdlibvars) linter. <br>
```
ccoVeille, line 584:
```suggestion
**Reason**: Cleaner code and meaningful constants instead of magical numbers/identifiers. This checker is similar to the [usestdlibvars](https://golangci-lint.run/usage/linters/#usestdlibvars) linter. <br>
```

ccoVeille: One typo+one suggestion

ccoVeille (approved):

ccoVeille (approved):

ccoVeille (approved):

Issue #47, opened 2024-01-05
feature: Detect Contains err.Error()


Limero:
Running `testifylint` on a large codebase at work I noticed that people sometimes use `Contains` to check error messages, something that `testifylint` currently doesn't detect

Expected results:
```go
❌   assert.Contains(t, err.Error(), "error")
     require.Contains(t, err.Error(), "error")
     s.Contains(err.Error(), "error")

✅   assert.ErrorContains(t, err, "error")
     require.ErrorContains(t, err, "error")
     s.ErrorContains(err, "error")
```

Antonboom:
Hi!

Thank you for feedback.
Related to https://github.com/Antonboom/testifylint/blob/master/CONTRIBUTING.md#error-compare

nickajacks1:
I see that `assert.ErrorContains(t, err, "not found")` is discouraged in favor of `ErrorIs`.  I agree with that guidance when it's an option, but sometimes errors are unexported, meaning you have no choice but to use `ErrorContains`.  I understand that checking the error string is a bit fragile, but what alternatives would you suggest? Perhaps simply `assert.Error`?

Some justification I can think of for using `assert.Error` alone and ignoring the content of the error is that if your actual code can't check the type of error that is returned, that means that the content of the error doesn't affect the control flow, and thus the unit tests shouldn't need to be coupled to the human-readable error messages.

ccoVeille:
For records, I started working on the implementation of this feature request.

https://github.com/Antonboom/testifylint/blob/master/CONTRIBUTING.md#error-compare

Stay tuned, and ping me if you see no progress in the next month.

ccoVeille:
I agree with @Limero @nickajacks1 there is a need for a checker that will look for detecting this.

```
assert.Contains(t, err.Error(), "error") require.Contains(t, err.Error(), "error") s.Contains(err.Error(), "error")
```

This should be a checker activated by default.

As @Antonboom said, there is a need for a checker these by .ErrorContains(t, err, ErrSentinel)

But these are two separate needs. While the first one is legitimate and should be activated by default. The second one is more opinionated and could lead to problems.

We have to consider ErrorContains could be used for wrapped error to check one word is present in the error message

Let's consider a code returning that

fmt.Errorf("error %w - code %s, ErrSentinel, "foo")

I would use 

assert.ErrorIs(t, err, ErrSentinel)

but then I would use

assert.ErrorContains(t, err, "foo")

These use cases are legitimate.


[...]

PR #162, opened 2024-06-30
doc: update error-compare and define unwrapped-error

mmorel-35:
Following the different exchanges about error compare, I believe testifylint shall be allowed to recommend the use of ErrorContains, ErrorEqual, ErrorIs and NotErrorIs.

See https://github.com/Antonboom/testifylint/issues/47#issuecomment-2047862514 for more context

I believe ErrorIs is only applicable to wrapped errors but I might be missing something.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -143,7 +143,7 @@ Describe a new checker in [checkers section](./README.md#checkers).
 - [require-len](#require-len)
 - [suite-test-name](#suite-test-name)
 - [useless-assert](#useless-assert)
-
+- [unwrapped-error](#unwrapped-error)
 ---
 
 ### elements-match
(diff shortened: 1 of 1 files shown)


Antonboom:
ErrorContains is anti-pattern and, I think, must be flagged by default. We can consider an option to turn off it

CONTRIBUTING.md
```diff
@@ -172,19 +173,38 @@ Describe a new checker in [checkers section](./README.md#checkers).
...
 
+✅   assert.ErrorContains(t, err, "not found")
+     assert.EqualError(t, err, "user not found")
+     assert.ErrorIs(t, err, ErrUserNotFound)
+     assert.NotErrorIs(t, err, errSentinel)
+```
+
+**Autofix**: true. <br>
+**Enabled by default**: true. <br>
+**Reason**: The `Error()` method on the `error` interface exists for humans, not code. <br>
```
ccoVeille, line 190: Using testify methods are error prone as they avoid nil pointer exception.

CONTRIBUTING.md
```diff
@@ -172,19 +173,38 @@ Describe a new checker in [checkers section](./README.md#checkers).
...
+
+
 ✅   assert.ErrorIs(t, err, ErrUserNotFound)
+     assert.ErrorAs(t, err, &target)
 ```
 
 **Autofix**: false. <br>
 **Enabled by default**: true. <br>
-**Reason**: The `Error()` method on the `error` interface exists for humans, not code. <br>
+**Reason**: The `assert.ErrorContains()` is an anti-pattern. The test needs to be adapted to use `assert.ErrorIs()` or `assert.ErrorAs()` methods. <br>
```
ccoVeille, line 207: I still disagree it's a bad pattern when used to check for example a value is provided to sprintf, like 

```go
fmt.Errorf("%w: %d", ErrUnexpectedStatusCode, res.Status)
```

+  

```go
assert.ErrorIs(t, err, ErrUnexpectedStatusCode)
assert.ErrorContains(t, err, "404")
```

But I know most people are simply not using the right way.


```suggestion
**Reason**: The `assert.ErrorContains()` is commonly used as an anti-pattern. The code and tests often needs to be adapted to be able to use `assert.ErrorIs()` or `assert.ErrorAs()` methods. <br>
```

CONTRIBUTING.md
```diff
@@ -172,19 +173,38 @@ Describe a new checker in [checkers section](./README.md#checkers).
...
-❌   assert.ErrorContains(t, err, "not found")
-     assert.EqualError(t, err, "user not found")
+❌   assert.Contains(t, err.Error(), "not found")
      assert.Equal(t, err.Error(), "user not found")
      assert.Equal(t, err, errSentinel) // Through `reflect.DeepEqual` causes error strings to be compared.
      assert.NotEqual(t, err, errSentinel)
      // etc.
 
+✅   assert.ErrorContains(t, err, "not found")
+     assert.EqualError(t, err, "user not found")
```
ccoVeille, line 183: I'm a bit afraid this one is valid

ccoVeille (approved):

CONTRIBUTING.md
ccoVeille: My point is that ErrorEqual should be part of the sane checker than ErrorContains.

It should be discouraged

ccoVeille: Maybe, let's wait for Anton's feedback

ccoVeille:
These names make more sense 👍

ccoVeille (approved):

CONTRIBUTING.md
ccoVeille: @Antonboom maybe you missed this old thread 

https://github.com/Antonboom/testifylint/pull/162/commits/4cb531e2e45e911b6ea77f3616b8c0e41dda42dd#r1661489366

You said ErrorContains is an antipattern, what is your point of view about ErrorEqual ?

Issue #158, opened 2024-06-26
Consider adding optional revive linters


ccoVeille:
I noticed than many of the code review I'm doing are about code simplification are about early-return pattern.

I recommend using revive that could detect this pattern. early-return is one we need I think

We may review a few more linters in the process but it would definitely help I think

_Originally posted by @ccoVeille in https://github.com/Antonboom/testifylint/pull/155#discussion_r1654060753_

ccoVeille:
I know, it's a pain. That's why I started working on a personal project to document such things and provide ready to go golangci-lint configuration files

https://github.com/ccoVeille/golangci-lint-config-examples

I will populate it in the next days

ccoVeille:
@mmorel-35 you can have a look at 

https://github.com/ccoVeille/golangci-lint-config-examples/tree/main/03-safe

PR #164, opened 2024-07-04
enable several rules from revive

mmorel-35:
Related to #158

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -67,6 +67,36 @@ linters-settings:
   lll:
     line-length: 130
 
+  revive:
+    rules:
+      - name: blank-imports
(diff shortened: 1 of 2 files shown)


ccoVeille:
We might need to enable more, but this one is great.
I didn't have a chance to keep working on https://github.com/ccoVeille/golangci-lint-config-examples

I think we could also enable

- empty-block
- if-return
- identical-branch
- indent-error-flow
- superfluous-else
- unreachable-code
- useless-break

All these are safe and works perfectly with early-return

ccoVeille (approved):
Thanks

ccoVeille (approved):

Antonboom (approved):

Issue #170, opened 2024-08-05
`formatter` lint rule only warns when importing `assert` package


pgimalac:
Here is a minimal reproducible example where the formatter rule should suggest not using `fmt.Sprintf`, but doesn't.
```go
package main

import (
	"fmt"
	"testing"

	"github.com/stretchr/testify/require"
)

func TestA(t *testing.T) {
	require.True(t, false, fmt.Sprintf("expected %v, got %v", true, false))
}
```

Adding a blank import of `github.com/stretchr/testify/assert` (or any assertion using it) will make the `formatter` rule complain.

ccoVeille:
Thanks for reporting this. I appreciate your feedback, and the minimum reproducible code.

We will have a look.

This one is indeed a bug, if confirmed.

The fix should be easy :tm: (famous last words 😅🤣)

pgimalac:
I took a look and I believe I found the issue, I'll validate and open a PR.

Antonboom:
@pgimalac thanks!

Fixed.

PR #171, opened 2024-08-06
Fix `formatter` when not importing assert package

pgimalac:
Fixes https://github.com/Antonboom/testifylint/issues/170.

I tested manually that this fixes the issue on the example provided in the issue.

I wasn't able to add a test as it would require a whole new test suite (the bug was only when not importing the `assert` package), let me know if I should try to do that.


Antonboom:
@pgimalac thanks for contribution

I added tests and other corner cases in separate PR 👌

Issue #166, opened 2024-07-30
len: Len instead of Equal when comparing two len's


Limero:
In a large codebase at work I noticed that people sometimes compare the values of two `len()` with `Equal`, something that `testifylint` v1.4.3 currently doesn't detect. It should suggest using `assert.Len()` the same as if you compare with a const, i.e. `assert.Equal(t, len(expected), 2)`

Expected results:
```go
expected := []string{"a", "b"}
actual := map[string]bool{
  "a": true,
  "b": true,
}

❌ assert.Equal(t, len(expected), len(actual))
✅ assert.Len(t, expected, len(actual))
```

ccoVeille:
Thanks for reporting the issue. It's about a false-negative, so it's a good news.

I'm unsure what testifylint should do here, we might have to discuss about it.

But here are some real life examples

https://github.com/cli/cli/blob/b05bddc210a1e45a25b76f9b320239e4c36f03ee/pkg/jsonfieldstest/jsonfieldstest.go#L45

https://github.com/uber/tchannel-go/blob/af146f5a54ac5787563a6857ed9b6ff26394f0ca/relay/relaytest/mock_stats.go#L146

https://github.com/bufbuild/buf/blob/3ef45f86e43fea376f93df6ae8c115d27637513e/private/pkg/storage/storagetesting/storagetesting.go#L121

https://github.com/netobserv/netobserv-ebpf-agent/blob/47e9fb8c3ef5f9a651d6e19c73a340ee5dec8520/e2e/basic/common.go#L121

At least the pattern is frequently used

PR #204, opened 2024-11-17
len: support len-len and value-len cases

Antonboom:
Closes https://github.com/Antonboom/testifylint/issues/166


README.md
```diff
@@ -618,19 +622,41 @@ P.S. Look at [testify's issue](https://github.com/stretchr/testify/issues/772),
...
+assert.Len(t, arr, value)
+assert.Len(t, arr, len(expArr))
 ```
 
 **Autofix**: true. <br>
 **Enabled by default**: true. <br>
 **Reason**: More appropriate `testify` API with clearer failure message.
 
+> [!CAUTION]
+> The checker ignores assertions in which length checking is not a priority, e.g.
```
ccoVeille, line 655: Maybe a link to the issue where the reason are provided could help people to understand it's normal, and it would be a bad thing to change

internal/checkers/len.go
```diff
@@ -1,65 +1,105 @@
...
 // and requires
 //
-//	assert.Len(t, arr, 3)
+//	assert.Len(t, arr, 42)
+//	assert.Len(t, arr, value)
+//	assert.Len(t, arr, len(expArr))
+//
+// The checker ignores assertions in which length checking is not a priority, e.g
+//
+//	assert.Equal(t, len(arr), value)
```
ccoVeille, line 39: Add the link here too

ccoVeille: I would add it in the code, but I'm unsure about the exact line

PR #212, opened 2025-01-04
fix: negative positive with type conversion

### LINKED ISSUE #210: negative-positive: bad suggested fix for `len` comparison
```
% cat my_test.go
package main

import (
        "testing"

        "github.com/stretchr/testify/assert"
)

func TestCommandsDontUseUnderscoreInName(t *testing.T) {
        x := []string{}
        assert.True(t, len(x) > 0)
}
```

```
% testifylint --fix my_test.go
/Users/denis.bilenko/work/cli/my_test.go:11:2: negative-positive: use assert.Positive
```

```
% diff -u my_test.go.bak my_test.go
--- my_test.go.bak      2025-01-02 11:47:19
+++ my_test.go  2025-01-02 11:49:03
@@ -8,5 +8,5 @@

 func TestCommandsDontUseUnderscoreInName(t *testing.T) {
        x := []string{}
-       assert.True(t, len(x) > 0)
+       assert.Positive(t, x)
 }
```

Expected `assert.NotEmpty(t, x)` or `assert.Positive(t, len(x))` but not the above.

ccoVeille:
The code was too naive, and was stripping any code with only one arg

Fixes #210


Antonboom:
Hi, @ccoVeille 

Len call is not type conversion and we have separate helper for it.

I expected:
- new type conversion test cases (with untyping support and proofs in debug package), if we have not yet it
- just len-case support

ccoVeille:
> Len call is not type conversion and we have separate helper for it.

I totally agree with you. What I don't get is that comment is irrelevant with the fix I'm doing here

I used Len in the test because len was reported by the person who reported the bug.

The problem is about avoiding suggesting to replace `assert.True(t, whatever(s) > 0)` to `assert.Positive(t, s)`.

The replacement is OK when it's a numerical conversion. But when whatever is `len`, `min`, `max`, `func (int) int`, anything with one argument returning something comparable with > 0

I hope it's clearer.

I agree with the need of adding test to helpers, but I found none for other helpers, assuming you were not looking for it. I decided to do not fight this dragon. Feel free to add them.

ccoVeille:
I'm just noticing that the rules simplification by removing the unnecessary type conversion should/could also be applied to other asserter, such as Zero, Empty, Positive, Negative, Len

```go
assert.Positive(t, int(42))
```

Also the type conversion suggestion is only applied when using the `-fix`, so maybe it could be part of a dedicated linter.

Another way to consider this PR is to move out the type conversion from this PR and move it to another rule that could

Issue #235, opened 2025-03-25
False-positive `useless-assert: meaningless assertion` with `require.Positive`


zak-pawel:
When I run `TestPositiveNumber`
```go
package main

import (
	"testing"

	"github.com/stretchr/testify/require"
)

func number() uint64 {
	return 0
}

func TestPositiveNumber(t *testing.T) {
	n := number()
	require.Positive(t, n)
}
```

it will fail:
```
--- FAIL: TestPositiveNumber (0.00s)
    positive_test.go:15: 
                Error Trace:    /home/pzak/work/positive/positive_test.go:15
                Error:          "0" is not positive%!(EXTRA uint64=0)
                Test:           TestPositiveNumber
FAIL
exit status 1
FAIL    positive        0.003s
```

But `useless-assert` tells me that this `require.Positive` is meaningless :(
```
➜  go install github.com/Antonboom/testifylint@latest
➜  testifylint -V=full                               
/home/pzak/go/bin/testifylint version devel comments-go-here buildID=c18b8ad775f31891c5df40363e114e646699b86fcaf86ab753ee1a8baee8fda5
➜  testifylint ./...                                 
/home/pzak/work/positive/positive_test.go:15:2: useless-assert: meaningless assertion
```

PR #236, opened 2025-03-25
fix: false-positive useless-assert on Positive function

semihbkgr:
It was reporting a false-positive `useless-assert` when calling the `Positive` function with `unsigned` integer types. It overlooked the fact that an unsigned integer might be zero, which is not positive. It is only useless to call `Negative` function with an `unsigned` integer.

See #235


Antonboom:
@semihbkgr hi!

thank you for contribution, but in the next time, please, do not ignore documentation changes and follow the TDD
thanks

PR #246, opened 2025-06-13
equal-values: fix false positive with any

ccoVeille:
When variables are both any, we should not report.

### DIFF
--- a/internal/checkers/equal_values.go
+++ b/internal/checkers/equal_values.go
@@ -50,7 +50,7 @@ func (checker EqualValues) Check(pass *analysis.Pass, call *CallMeta) *analysis.
 	// Type of one of arguments is equivalent to any.
 	if isEmptyInterfaceType(ft) || isEmptyInterfaceType(st) {
 		// EqualValues is ok here.
-		// Equal would check their types and would fail
+		// Equal would check their types and would fail.
 		return nil
 	}
 


internal/checkers/equal_values.go
```diff
@@ -43,9 +43,17 @@ func (checker EqualValues) Check(pass *analysis.Pass, call *CallMeta) *analysis.
...
 	ft, st := pass.TypesInfo.TypeOf(first), pass.TypesInfo.TypeOf(second)
-	if types.Identical(ft, st) {
-		proposed := strings.TrimSuffix(assrn, "Values")
-		return newUseFunctionDiagnostic(checker.Name(), call, proposed)
+	if !types.Identical(ft, st) {
+		return nil
 	}
-	return nil
+
+	// one argument type is equivalent to any
```
Antonboom, line 50:
```suggestion
	// Type of one of arguments is equivalent to any.
```

internal/checkers/equal_values.go
```diff
@@ -43,9 +43,17 @@ func (checker EqualValues) Check(pass *analysis.Pass, call *CallMeta) *analysis.
...
+	if isEmptyInterfaceType(ft) || isEmptyInterfaceType(st) {
+		// EqualValues is ok here.
+		// Equal would check their types and would fail
```
Antonboom, line 53:
```suggestion
		// Equal would check their types and would fail.
```

Antonboom (approved):

PR #247, opened 2025-06-16
suite_dont_use_pkg: no testing.T with suite

HaraldNordgren:
Create checker to make sure that `suite` methods are not passing a bare `testing.T`.


HaraldNordgren:
Ping @Antonboom, can I get some review here please? 🤗

Antonboom:
Hi, @HaraldNordgren 

Thank you for contribution!

Looks like improvement of [#suite-method-signature](https://github.com/Antonboom/testifylint?tab=readme-ov-file#suite-method-signature) or [#suite-dont-use-pkg](https://github.com/Antonboom/testifylint?tab=readme-ov-file#suite-dont-use-pkg) and not a separate checker.

In the next time I recommend to make issue first to avoid the extra work 🙂

What are your thoughts after taking into account the logic of the two checkers above?

HaraldNordgren:
Hi @Antonboom! Sorry I missed this, I changed my GitHub email routing and lost my email notifications for this thread 😬

I would suggest we put it in `suite-dont-use-pkg` primarily because that one supports auto-fixing. Is it in any way a concern for you that one of these two previous checkers become too bloated after adding this functionality?

HaraldNordgren:
@Antonboom Updated, can you please take another look? 🤗

HaraldNordgren:
@Antonboom Updated, can you please take another look? 🤗

HaraldNordgren:
@Antonboom Updated, can you please take another look? 🤗

HaraldNordgren:
@Antonboom Updated, can you please take another look? 🤗

HaraldNordgren:
Hi @Antonboom is there any chance of getting this merged? If so, I will take a look at resolving the merge conflicts here.

PR #257, opened 2025-08-16
new checker 'mock-expect'

### LINKED ISSUE #250: mock-expect: suggest using EXPECT
Since `testify/mock` is used for `.On()` calls, the linter could be part of testifylint. I'll try to open a PR.
Thanks for taking a look.

_Originally posted by @d0ubletr0uble in https://github.com/golangci/golangci-lint/pull/5926#issuecomment-3063152720_

d0ubletr0uble:
Closes https://github.com/Antonboom/testifylint/issues/250


Antonboom:
Hi, @d0ubletr0uble 

Thank you for contributing!

Maybe you can suggest some large code bases with using of mockery?
(to test the checker)

P.S. A am personally a fan of https://github.com/uber-go/mock and not familiar with https://github.com/vektra/mockery

d0ubletr0uble:
https://github.com/kyma-project/eventing-manager is a good candidate to test the linter.

```
git clone --depth=1 https://github.com/kyma-project/eventing-manager
cd eventing-manager

testifylint -disable-all -enable mock-expect ./...
testifylint -disable-all -enable mock-expect -fix ./...
```
It even detects cases where `.GetSecret()` and `.Stop()` methods break interface contract since you can pass anything to `.On()` method. Such mocks are registered but will never be called.

Linter does break one `.PatchApply(...).Run(...)` call where `.Run(...)` argument types need to be updated. Will see if I can detect such cases and skip autofix for them.

Also, it currently ignores [gingo](https://github.com/onsi/ginkgo) test suites. I'll check what can be done about that.

d0ubletr0uble:
Added [gingo](https://github.com/onsi/ginkgo) support by dropping restriction to check only `Test` functions.

https://github.com/pluralsh/console is a code base that uses `gingo` test suites and `mockery`.

```
git clone --depth=1 https://github.com/pluralsh/console
cd console/go/controller

testifylint -disable-all -enable mock-expect ./...
testifylint -disable-all -enable mock-expect -fix ./...
```

It does break cases where too many arguments were passed into original `.On()` function call.

d0ubletr0uble:
Made the linter to skip calls with `.Run(...)` in the chain. Fixes the issue where it would mess up the arguments.

thespags:
Hi. I'm curious as to the status. This would be great to get into testifylint.

d0ubletr0uble:
All done on my end, just waiting for feedback or a merge. You can use the fork in the meantime.

Issue #240, opened 2025-05-19
error_nil/error_is_as: IsType vs nil/error


ccoVeille:
By looking of code using IsType, I found bad code that are not linted yet.

https://github.com/PacktPublishing/Hands-On-Serverless-Applications-with-Go/blob/f1701bc78e92fd971246fd4bced7c7bddbf22fec/ch10/findAll/main_test.go#L17

```go
assert.IsType(t, nil, err)
```

=> assert.NoError should be considered

https://github.com/Infomaniak/kchat-webapp/blob/932ec3af126f6225f9424624dad54127ac24f7ca/server/channels/store/storetest/desktop_tokens_store.go#L36

```go
assert.IsType(t, &store.ErrNotFound{}, err)
```

=> ErrorAs should be considered

Issue #190, opened 2024-10-03
error-is-as: support `ErrorIs(nil, err)`


Antonboom:
https://github.com/search?q=language%3Ago+%22assert.ErrorIs%28t%2C+nil%22&type=code

PR #260, opened 2025-09-04
error-is-as: error-nil: support IsType and IsNotType

Antonboom:
Closes https://github.com/Antonboom/testifylint/issues/190
Closes https://github.com/Antonboom/testifylint/issues/240


Antonboom:
@ccoVeille thank you very much!

for the history: no full fix for `IsType`/`IsNotType` in error-is-as, because a lot of complex (multi-line) cases are possible here

Issue #175, opened 2024-09-25
Version flag unsuported


Xarakiri:
Installed last version of testifylint
```
go install github.com/Antonboom/testifylint@latest
```

And got an error while trying to print version
```
➜  ~ testifylint -V
testifylint: unsupported flag value: -V=true (use -V=full)
➜  ~
```

`-V=true` and `-V=full` also don't work
```
➜  ~ testifylint -V=true
testifylint: unsupported flag value: -V=true (use -V=full)
➜  ~ testifylint -V=full
/Users/mac/go/bin/testifylint version devel comments-go-here buildID=d96eb5ef14112910370b26b280f200314c2c46316154bed8163c70903afd903e
➜  ~
```

I expect got something like this:
```
➜  ~ testifylint -V
testifylint 1.4.3
```

Antonboom:
```bash
$ go run github.com/Antonboom/testifylint@latest -V
version unknown built with go1.25.0 from unknown on unknown
```

```bash
$ ~/Downloads/testifylint_1.6.3_darwin_arm64/testifylint -V
version 1.6.3 built with go1.25.1 from db60550f0d50d4b52d9450dd7ae81cb36c436334 on 2025-09-10T15:29:32Z
```

PR #263, opened 2025-09-10
Version support

Antonboom:
Closes https://github.com/Antonboom/testifylint/issues/175

```go
$  testifylint -V
version unknown built with go1.25.0 from unknown on unknown
```


Antonboom:
Approach is fixed in https://github.com/Antonboom/testifylint/commit/7857840997deaa38a73e5c725dcb6109dac0cf6d

PR #291, opened 2026-03-13
feat: new checkers `http-method` and `http-status-code`

### LINKED ISSUE #141: http-const: suggest `http` package constants in HTTP asserts

mmorel-35:
Splits the `http-const` checker into two focused checkers aligned with [`usestdlibvars`](https://github.com/sashamelentyev/usestdlibvars) naming conventions. No backwards compatibility required.

### New checkers

- **`http-method`** — flags HTTP method string literals in all testify HTTP assertions (`HTTPStatusCode`, `HTTPBody*`, `HTTPError`, `HTTPRedirect`, `HTTPSuccess`)
- **`http-status-code`** — flags HTTP status code integer literals in `HTTPStatusCode`

```go
// ❌ Before (both violations caught by old http-const)
assert.HTTPStatusCode(t, handler, "GET", "/", nil, 200)

// ✅ After
assert.HTTPStatusCode(t, handler, http.MethodGet, "/", nil, http.StatusOK)
//                                 ^http-method^       ^http-status-code^
```

### Changes

- **Deleted** `http_const.go` and `gen_http_const.go`; replaced with `http_method.go`, `http_status_code.go` and corresponding testgen files
- **Moved** shared helpers (`newHTTPMethodTextEdit`, `newHTTPStatusCodeTextEdit`, `httpQualifiedName`) into `helpers_http.go`
- **Updated** registry, registry tests, and factory tests to reference the two new checkers
- **Regenerated** test data and added dedicated `http-method-no-import` / `http-status-code-no-import` test cases (auto-import fix coverage)
- **Updated** README table and checker documentation sections

Covers #155
Closes #141


ccoVeille (approved):
👍

But I would have separated http.Status* and http.Method*

I remember a discussion where @alexandear was involved in go github where people where disliking the http.Method*, but where OK for http.Status* one

Do you remember the context @alexandear ?

alexandear:
> 👍
> 
> But I would have separated http.Status* and http.Method*
> 
> Do you remember the context @alexandear ?

From Brad Fitzpatrick https://github.com/perkeep/perkeep/pull/1649#issuecomment-1400790540:

> I'd rather not do this change.
> 
> I added the http.Method* constants to Go but have regretted it ever since. They're worse for readers than "GET" or "POST", as the http.MethodGet and http.MethodPost share such a  long http.Method prefix that they blend together.
>
> Please use the strings instead.
> 
> As for the Status constants, those are fine, but even "200" and "404" and "500" are common enough that I don't mind them. But definitely use constants for anything that isn't 200, 404, or 500.

mmorel-35:
Notice that the AddImports method here is reused in several other PR. This version shall be used by them as it is the most evolved version yet .

internal/analysisutil/file.go
```diff
@@ -2,10 +2,25 @@ package analysisutil
...
 	"slices"
 	"strconv"
 )
 
+// pkgBaseName returns the default package name for the given import path.
+// It handles versioned module paths (e.g., "example.com/pkg/v2" → "pkg").
+func pkgBaseName(importPath string) string {
+	base := path.Base(importPath)
+	// If the last element looks like a major version tag (v2, v3, ...), use the preceding element.
+	if len(base) > 1 && base[0] == 'v' {
```
ccoVeille: strings.HasPrefix(base, "v") would be simpler I think

ccoVeille: But then based on the fact you remove the first character, consider using strings.CutPrefix that will do the detection and cleaning

mmorel-35: I enabled modernize in https://github.com/Antonboom/testifylint/pull/306

It is usually able to catch those patterns

internal/analysisutil/file.go
```diff
@@ -2,10 +2,28 @@ package analysisutil
...
 	"strconv"
+	"strings"
 )
 
+// pkgBaseName returns the default package name for the given import path.
+// It handles versioned module paths (e.g., "example.com/pkg/v2" → "pkg").
+func pkgBaseName(importPath string) string {
+	base := path.Base(importPath)
+	// If the last element looks like a major version tag (v2, v3, ...), use the preceding element.
+	if len(base) > 1 {
```
ccoVeille: This if statement is not needed

internal/analysisutil/file.go
```diff
@@ -2,10 +2,28 @@ package analysisutil
...
+		if after, found := strings.CutPrefix(base, "v"); found {
```
ccoVeille: I would move to this


```suggestion
		after, found := strings.CutPrefix(base, "v")
		if ! found {
			return base
		}
```

Then no if branching is needed

ccoVeille:
> 👍
> 
> But I would have separated http.Status* and http.Method*
> 
> I remember a discussion where @alexandear was involved in go github where people where disliking the http.Method*, but where OK for http.Status* one
> 
> Do you remember the context @alexandear ?

>> From Brad Fitzpatrick https://github.com/perkeep/perkeep/pull/1649#issuecomment-1400790540:
> 
> > I'd rather not do this change.
> > 
> > I added the http.Method* constants to Go but have regretted it ever since. They're worse for readers than "GET" or "POST", as the http.MethodGet and http.MethodPost share such a  long http.Method prefix that they blend together.
> >
> > Please use the strings instead.
> > 
> > As for the Status constants, those are fine, but even "200" and "404" and "500" are common enough that I don't mind them. But definitely use constants for anything that isn't 200, 404, or 500.
> 
> 


@mmorel-35 what are your thoughts about this

https://github.com/sashamelentyev/usestdlibvars has 2 rules for this

mmorel-35:
I believe it's better to be aligned with what other linters are already doing so in that case, let's split this in two checkers

ccoVeille (approved):

internal/checkers/helpers_import.go
```diff
@@ -0,0 +1,282 @@
...
+			switch imp.Name.Name {
+			case ".":
+				return "", nil, true // dot-import: symbols accessible without qualifier
+			case "_":
+				return "", nil, false // blank import: package not accessible
+			default:
+				return imp.Name.Name, nil, true // aliased import
+			}
+		}
+		return importBaseName(pkgPath), nil, true // regular import
```
ccoVeille: Not picking

I tend to leave early. Code is easier to read


```suggestion
		if imp.Name == nil {
			return importBaseName(pkgPath), nil, true // regular import
		}

		switch imp.Name.Name {
		case ".":
			return "", nil, true // dot-import: symbols accessible without qualifier
		case "_":
			return "", nil, false // blank import: package not accessible
		default:
			return imp.Name.Name, nil, true // aliased import
		}
```

internal/checkers/helpers_import.go
```diff
@@ -0,0 +1,282 @@
...
+// importBaseName returns the default local name for the given import path.
+// It handles versioned module paths (e.g., "example.com/pkg/v2" → "pkg").
+func importBaseName(importPath string) string {
+	base := path.Base(importPath)
+	if len(base) > 1 && base[0] == 'v' {
+		if _, err := strconv.Atoi(base[1:]); err == nil {
+			base = path.Base(path.Dir(importPath))
+		}
+	}
+	return base
```
ccoVeille: strings.CutPrefix

internal/checkers/helpers_import.go
```diff
@@ -0,0 +1,282 @@
...
+
+// isStdlibPath reports whether pkgPath belongs to the Go standard library.
+// A standard library package has no dot in its first path segment.
+func isStdlibPath(pkgPath string) bool {
+	slash := strings.IndexByte(pkgPath, '/')
+	if slash < 0 {
+		slash = len(pkgPath)
+	}
+	return !strings.Contains(pkgPath[:slash], ".")
+}
```
ccoVeille, line 85: "It's too dangerous to go alone, take this!"

```suggestion
func isStdlibPath(pkgPath string) bool {
// Split into at most 2 parts at the first slash
    firstSegment, _, _ := strings.Cut(pkgPath, "/")
    return !strings.Contains(firstSegment, ".")
}
```

internal/checkers/helpers_import.go
```diff
@@ -0,0 +1,282 @@
...
+		if err != nil || p == excludePath {
+			continue
+		}
+		if imp.Name != nil {
+			if imp.Name.Name != "_" && imp.Name.Name != "." {
+				names[imp.Name.Name] = struct{}{}
+			}
+		} else {
+			names[importBaseName(p)] = struct{}{}
+		}
```
ccoVeille: It could be clearer like this maybe


```suggestion
		var importName string
		if imp.Name != nil && imp.Name.Name != "_" && imp.Name.Name != "." {
				importName = imp.Name.Name
		} else {
			importName = importBaseName(p)
		}
		names[importName] = struct{}{}
```

internal/checkers/helpers_import.go
mmorel-35: That shall be warned by revive or gocritic, why isn't it ?

PR #292, opened 2026-03-13
regexp: warn when rx argument is not string or *regexp.Regexp

### LINKED ISSUE #81: Add type checks for `assert.Regexp` regexp argument
The signature for [`assert.Regexp`](https://pkg.go.dev/github.com/stretchr/testify/assert#Regexp) (and associated methods) is:

```go
func Regexp(t TestingT, rx interface{}, str interface{}, msgAndArgs ...interface{}) bool
```

However `rx` should only be `*regexp.Regexp` or `string`.

Related: stretchr/testify#1585
Cc: @kevinburkesegment

mmorel-35:
`assert.Regexp`/`assert.NotRegexp` accept `rx interface{}`, but testify's `matchRegexp` only handles `*regexp.Regexp` and `string` correctly — anything else silently goes through `regexp.MustCompile(fmt.Sprint(rx))`, which produces unexpected results (e.g. `[]byte("pattern")` compiles to its decimal byte representation).

## Changes

- **`internal/checkers/regexp.go`**: Extended `Check()` to emit `"use string or *regexp.Regexp as the first argument"` when `rx` is neither a string type (typed, untyped, or string-underlying alias) nor `*regexp.Regexp`. Added `isStringOrRegexpType` helper.
- **`internal/testgen/gen_regexp.go`**: Added invalid test cases (`[]byte` arg) and valid test cases (pre-compiled `*regexp.Regexp` variable); updated template to declare `compiledRegexp`.

```go
// Now flagged:
assert.Regexp(t, []byte(`\w+`), output) // regexp: use string or *regexp.Regexp as the first argument

// Valid — no warning:
assert.Regexp(t, `\w+`, output)
assert.Regexp(t, compiledRegexp, output)
```

Closes #81

### DIFF
--- a/internal/checkers/regexp.go
+++ b/internal/checkers/regexp.go
@@ -2,6 +2,7 @@ package checkers
 
 import (
 	"go/ast"
+	"go/types"
 
 	"golang.org/x/tools/go/analysis"
 )
@@ -15,6 +16,14 @@ import (
 //
 //	assert.Regexp(t, `\[.*\] DEBUG \(.*TestNew.*\): message`, out)
 //	assert.NotRegexp(t, `\[.*\] TRACE message`, out)
+//
+// Also detects situations like
+//
+//	assert.Regexp(t, []byte(`\w+`), str)
+//
+// and requires
+//
+//	assert.Regexp(t, `\w+`, str) // or *regexp.Regexp
 type Regexp struct{}
 
 // NewRegexp constructs Regexp checker.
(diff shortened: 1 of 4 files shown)


ccoVeille (approved):

internal/checkers/regexp.go
```diff
@@ -32,13 +41,50 @@ func (checker Regexp) Check(pass *analysis.Pass, call *CallMeta) *analysis.Diagn
...
+		return false
+	}
+
+	// Conservatively accept type parameters to avoid false positives in generic code.
+	if _, ok := types.Unalias(t).(*types.TypeParam); ok {
+		return true
+	}
+
+	// Check for string types (includes string, untyped string, and string-underlying type aliases).
+	if isUnderlying(pass, e, types.IsString) {
```
dolmen: assert.Regexp must NOT accept any string type. Because we want to prevent the case where the user swapped parameters.

We must only accept:
* strict `string`
* `*regexp.Regexp`
* alias to `string` or `*regexp.Regexp`

internal/checkers/regexp.go
```diff
@@ -32,13 +41,50 @@ func (checker Regexp) Check(pass *analysis.Pass, call *CallMeta) *analysis.Diagn
...
+// type (including type aliases), or a type parameter (accepted conservatively to avoid false
+// positives in generic code).
+func isStringOrRegexpType(pass *analysis.Pass, e ast.Expr) bool {
+	t := pass.TypesInfo.TypeOf(e)
+	if t == nil {
+		return false
+	}
+
+	// Conservatively accept type parameters to avoid false positives in generic code.
+	if _, ok := types.Unalias(t).(*types.TypeParam); ok {
```
dolmen: Don't drop the result of unaliasing: instead that's what we should manipulate from this point.

internal/checkers/regexp.go
```diff
@@ -32,13 +41,50 @@ func (checker Regexp) Check(pass *analysis.Pass, call *CallMeta) *analysis.Diagn
...
+// (including untyped strings, defined string types, and string-based type aliases), *regexp.Regexp
+// type (including type aliases), or a type parameter (accepted conservatively to avoid false
+// positives in generic code).
+func isStringOrRegexpType(pass *analysis.Pass, e ast.Expr) bool {
+	t := pass.TypesInfo.TypeOf(e)
+	if t == nil {
+		return false
+	}
+
+	// Conservatively accept type parameters to avoid false positives in generic code.
```
dolmen: It is not clear to me what you are refering to related to generic code here.
Add here, in the comment, a concrete example.

I don't see [`*types.TypeParam`](https://pkg.go.dev/go/types#TypeParam) handled. If that's what you were thinking about, that is different from [`*types.Alias`](https://pkg.go.dev/go/types#Alias), and that must be handled separately.

internal/checkers/regexp.go
```diff
@@ -32,13 +41,50 @@ func (checker Regexp) Check(pass *analysis.Pass, call *CallMeta) *analysis.Diagn
...
+		return true
+	}
+
+	// Check for *regexp.Regexp (using Unalias on both the type and pointer element to handle
+	// type aliases such as `type MyRx = *regexp.Regexp` or `type MyRegexp = regexp.Regexp`).
+	ptr, ok := types.Unalias(t).(*types.Pointer)
+	if !ok {
+		return false
+	}
+	named, ok := types.Unalias(ptr.Elem()).(*types.Named)
```
dolmen: I don't see a reason for accepting aliases to `regexp.Regexp` here.

dolmen (changes requested):
We must only accept:
* any `*types.TypeParam`
* `*types.Basic` that is strict `string` (BasicKind is `String` or `UntypedString`) (but NOT anything that is `string` underlying)
* any `*types.Alias` to above

The current implementation is too lax and doesn't handle TypeParam.

internal/testgen/gen_regexp.go
```diff
@@ -15,30 +15,69 @@ func (RegexpTestsGenerator) Checker() checkers.Checker {
...
+				Fn:         "Regexp",
+				Argsf:      "rxRegexpStructAlias, out",
+				ReportMsgf: reportInvalidArg,
+			},
+			{
+				Fn:         "NotRegexp",
+				Argsf:      "rxRegexpStructAlias, out",
+				ReportMsgf: reportInvalidArg,
 			},
 		},
```
ccoVeille, line 121: Can you add the following to tests

- using 42 as an integer (typed or untyped, signed or not
- using float
- stringer

All of these rely on the fmt.Sprint that is done with non string or *regexp.Regexp

```go
type foo struct{}

func (foo) String() string {
	return "42"
}

func TestFoo(t *testing.T) {
  assert.NotRegexp(t, false, "false")
	assert.NotRegexp(t, 42, "42")
	assert.NotRegexp(t, 42.0, "42")
	assert.NotRegexp(t, int8(42), "42")
	assert.NotRegexp(t, foo{}, "42")
}
```

All of these relies on something that would be better to avoid.

https://goplay.tools/snippet/Mmo4V36BHD1

Maybe something specific to stringer could be considered as a stringer can be rewritten to use .String()

PR #295, opened 2026-03-13
Fix false positives on `collect.Errorf` calls in `require-error` checker

### LINKED ISSUE #287: False positives on collect.Errorf calls
```go
type checker struct {
	foo  []any
	bar  []any
	name string
}

func (c *checker) cond(collect *assert.CollectT) {
	if c.foo == nil && c.bar == nil {
		collect.Errorf("precondition not satisfied")
	}
	require.NotEmpty(collect, c.name)
}
```

`testifylint` reports:

```
require-error: for error assertions use require
```

at `collect.Errorf`.

`collect.Errorf` reports an error and isn't an error assertion.

Interestingly, it's only reported when there's another testify call in the same scope. 
If you remove the `require.NotEmpty` line, it's not triggered for the `collect.Errorf` call.

mmorel-35:
`*assert.CollectT` implements `testing.T` via its `Errorf(format string, args ...interface{})` method. The `require-error` checker was incorrectly matching this against the `Error`/`Errorf` assertion pattern, producing a false positive whenever another testify call appeared in the same scope.

## Changes

- **`internal/checkers/require_error.go`**: For non-package (method) calls, guard with `isAssertionsReceiverMethod` — a new helper that inspects the method's declared receiver type (name + package path) to confirm it is `*assert.Assertions` or `*require.Assertions`, not an incidental type like `*assert.CollectT`.
- **`analyzer/testdata/src/require-error-skip-logic/issue287_test.go`**: Regression test reproducing the exact scenario — `collect.Errorf` must not be flagged when paired with other testify calls in the same function.

## Example

```go
func (c *checker) cond(collect *assert.CollectT) {
    if c.foo == nil && c.bar == nil {
        collect.Errorf("precondition not satisfied") // no longer a false positive
    }
    require.NotEmpty(collect, c.name)
}
```

The `formatter` checker, which legitimately inspects `collect.Errorf` for format-string correctness, is unaffected.

Closes #287

### DIFF
--- a/analyzer/testdata/src/require-error-skip-logic/issue287_test.go
+++ b/analyzer/testdata/src/require-error-skip-logic/issue287_test.go
@@ -1,0 +1,20 @@
+package requireerrorskiplogic
+
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type checkerIssue287 struct {
+	foo  []any
+	bar  []any
+	name string
+}
+
+// collect.Errorf is not a testify assertion; it should not be reported.
+func (c *checkerIssue287) cond(collect *assert.CollectT) {
+	if c.foo == nil && c.bar == nil {
+		collect.Errorf("precondition not satisfied")
+	}
+	require.NotEmpty(collect, c.name)
+}
--- a/internal/checkers/require_error.go
+++ b/internal/checkers/require_error.go
@@ -2,10 +2,14 @@ package checkers
 
 import (
 	"go/ast"
+	"go/types"
 	"regexp"
 
 	"golang.org/x/tools/go/analysis"
 	"golang.org/x/tools/go/ast/inspector"
+
+	"github.com/Antonboom/testifylint/internal/analysisutil"
+	"github.com/Antonboom/testifylint/internal/testify"
 )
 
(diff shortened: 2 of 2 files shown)


ccoVeille (approved):

analyzer/testdata/src/require-error-skip-logic/issue287_test.go
```diff
@@ -0,0 +1,20 @@
+package requireerrorskiplogic
```
Antonboom, line 1: you didn't include the test into running list, lol

where is TDD

internal/checkers/require_error.go
```diff
@@ -247,3 +258,25 @@ type callMeta struct {
 func isNoErrorAssertion(fnName string) bool {
 	return (fnName == "NoError") || (fnName == "NoErrorf")
 }
+
+// isAssertionsReceiverMethod returns true if the method call is defined on *assert.Assertions
+// or *require.Assertions (including promoted methods on suite types), as opposed to other
+// types in those packages such as *assert.CollectT.
+func isAssertionsReceiverMethod(call *CallMeta) bool {
```
Antonboom, line 265: conflict with `NewCallMeta` logic, should be combined/reused/refactored

Antonboom (changes requested):

PR #298, opened 2026-03-15
Remove `t` argument only for pkg-level assertions

### LINKED ISSUE #277: panic: index out of range in checker error-is-as
I have encountered a panic in checker error-is-as:
```
$ testifylint --disable-all --enable=error-is-as ./...
panic: runtime error: index out of range [1] with length 1

goroutine 76 [running]:
github.com/Antonboom/testifylint/internal/checkers.ErrorIsAs.Check({}, 0xc0002460e0, 0xc00078a210)
        github.com/Antonboom/testifylint@v1.6.4/internal/checkers/error_is_as.go:58 +0x1025
github.com/Antonboom/testifylint/analyzer.(*testifyLint).regularCheck(0xc000629c50, 0xc0002460e0, 0xc000440bc8?)
        github.com/Antonboom/testifylint@v1.6.4/analyzer/analyzer.go:85 +0xa3
github.com/Antonboom/testifylint/analyzer.(*testifyLint).run.func1({0x6028fd0?, 0xc000539e40?})
        github.com/Antonboom/testifylint@v1.6.4/analyzer/analyzer.go:65 +0x38
golang.org/x/tools/go/ast/inspector.(*Inspector).Preorder(0xc000240030, {0xc000440bc8?, 0x0?, 0x0?}, 0xc000629bb0)
        golang.org/x/tools@v0.36.0/go/ast/inspector/inspector.go:123 +0x91
github.com/Antonboom/testifylint/analyzer.(*testifyLint).run(0xc000629c50, 0xc0002460e0)
        github.com/Antonboom/testifylint@v1.6.4/analyzer/analyzer.go:64 +0xcd
github.com/Antonboom/testifylint/analyzer.New.func1(0xc0002460e0)
        github.com/Antonboom/testifylint@v1.6.4/analyzer/analyzer.go:41 +0xfe
golang.org/x/tools/go/analysis/checker.(*Action).execOnce.func3(...)
        golang.org/x/tools@v0.36.0/go/analysis/checker/checker.go:367
golang.org/x/tools/go/analysis/checker.(*Action).execOnce(0xc0000b6840)
        golang.org/x/tools@v0.36.0/go/analysis/checker/checker.go:388 +0xb48
sync.(*Once).doSlow(0xc0002ac5c0?, 0x5e38420?)
        sync/once.go:78 +0xac
sync.(*Once).Do(...)
        sync/once.go:69
golang.org/x/tools/go/analysis/checker.(*Action).exec(...)
        golang.org/x/tools@v0.36.0/go/analysis/checker/checker.go:265
golang.org/x/tools/go/analysis/checker.execAll.func1(0xc00051d798?)
        golang.org/x/tools@v0.36.0/go/analysis/checker/checker.go:253 +0x45
created by golang.org/x/tools/go/analysis/checker.execAll in goroutine 1
        golang.org/x/tools@v0.36.0/go/analysis/checker/checker.go:259 +0x147
```

The following test file as input triggers the panic:
```
package util

import (
	"testing"

	"github.com/stretchr/testify/assert"
)

type a struct{}

func (a) Errorf(format string, args ...any) {}

func TestIsType(t *testing.T) {
	as := assert.New(t)
	as.IsType(&a{}, &a{})
}
```

What happens:

1. An assert instance is created with `New` such that the first argument to `IsType` is one of the objects to compare and not the `testing.T` instance.
2. The type of the first argument implements `assert.TestingT` (or `require.TestingT`) such that it is removed from `Args` by trimTArg [here](https://github.com/Antonboom/testifylint/blob/e42082070d4ba3ea360871475a78ca16bc466d1c/internal/checkers/call_meta.go#L119). This removal of the first "real" argument looks like a bug in itself, but is only a precondition for the panic.
3. With the first of the two arguments removed, there is only one element left in `call.Args` [here](https://github.com/Antonboom/testifylint/blob/e42082070d4ba3ea360871475a78ca16bc466d1c/internal/checkers/error_is_as.go#L58). The condition of the `if` is missing parenthesis around the "or" part. That causes the `isError(pass, call.Args[1])` to be executed even if `len(call.Args)` is less than two, and that causes the index-out-of-range panic.

`IsNotType` has the same problem [here](https://github.com/Antonboom/testifylint/blob/e42082070d4ba3ea360871475a78ca16bc466d1c/internal/checkers/error_is_as.go#L64).

mmorel-35:
\+ fix parentheses bug in `error-is-as`

Closes #277


analyzer/testdata/src/error-is-as-issue277/issue277_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+func (a) Errorf(format string, args ...any) {}
+
+func TestIssue277(t *testing.T) {
+	as := assert.New(t)
+
+	// These should not panic (assert.New(t) calls don't pass t as first arg).
+	// With assert.New(), the first arg to IsType is the "expected" value,
+	// but since type a implements TestingT, trimTArg removes it, leaving only 1 arg.
+	as.IsType(&a{}, &a{})
```
Antonboom, line 18: where is `// want "error-is-as: use assert\\.ErrorIs or assert\\.ErrorAs depending on the case"`?

analyzer/testdata/src/error-is-as-issue277/issue277_test.go
```diff
@@ -0,0 +1,23 @@
+package errorisasissue277
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// a implements assert.TestingT (has Errorf method), so trimTArg will remove it
```
Antonboom: it is not true, comment is bullshit

analyzer/testdata/src/error-is-as-issue277/issue277_test.go
```diff
@@ -0,0 +1,23 @@
...
+// from Args when used as first argument to IsType/IsNotType via assert.New(t).
+type a struct{}
+
+func (a) Errorf(format string, args ...any) {}
+
+func TestIssue277(t *testing.T) {
+	as := assert.New(t)
+
+	// These should not panic (assert.New(t) calls don't pass t as first arg).
```
Antonboom: remove useless comments, please

internal/checkers/error_is_as.go
```diff
@@ -55,13 +55,13 @@ func (checker ErrorIsAs) Check(pass *analysis.Pass, call *CallMeta) *analysis.Di
 		}
 
 	case "IsType":
-		if len(call.Args) >= 2 && isError(pass, call.Args[0]) || isError(pass, call.Args[1]) {
+		if len(call.Args) >= 2 && (isError(pass, call.Args[0]) || isError(pass, call.Args[1])) {
```
Antonboom, line 58: where is no similar gotchas in the project?

Antonboom (changes requested):

internal/checkers/error_is_as.go
Antonboom: checked, there are not

analyzer/testdata/src/error-is-as-issue277/issue277_test.go
Antonboom: not related to issue, my bad

analyzer/testdata/src/error-is-as-issue277/issue277_test.go
Antonboom: my apologies, confused with other interface

comment is true

PR #248, opened 2025-06-18
encoded-compare: fix suggestions for JSONEq and YAMLEq

ccoVeille:
I spotted it by using the following code:

```go    
var foo json.RawMessage
assert.EqualValues(t, `{"foo": "bar"}`, foo)
```

that testifylint suggested fixing like this:

```go
var foo json.RawMessage
assert.JSONEq(t, `{"foo": "bar"}`, foo)
```

It's invalid because JSONEq expects strings.

Later, I noticed the code was also suggesting to replace (even if it's invalid):

```go
var i int
assert.Equal(t, `{"foo": "bar"}`, i)
```

with:

```go
var i int
assert.JSONEq(t, `{"foo": "bar"}`, i)
```

The PR is divided in 2 commits to help to reproduce the issue:

- **encoded-compare: add tests to reveal false positives**
- **encoded-compare: fix suggestions for JSONEq and YAMLEq**

PR #301, opened 2026-03-15
encoded-compare: fix false positives for non-string types and JSON-named constants

### LINKED ISSUE #274: autofixing assert.Equal -> assert.JSONEq can break working tests
Running the autofix via golangci-lint will convert this code:

```go
func TestInvalidJSON(t *testing.T) {
	invalidJSON := []byte(`{invalid json`)
	results, err := Explode(&http.Request{}, invalidJSON)
	require.NoError(t, err, "should not return error for invalid JSON, just return original")
	require.Len(t, results, 1, "should return original bytes when JSON is invalid")
	assert.Equal(t, string(invalidJSON), string(results[0]), "should return original bytes unchanged")
}
```

to

```go
func TestInvalidJSON(t *testing.T) {
	invalidJSON := []byte(`{invalid json`)
	results, err := Explode(&http.Request{}, invalidJSON)
	require.NoError(t, err, "should not return error for invalid JSON, just return original")
	require.Len(t, results, 1, "should return original bytes when JSON is invalid")
	assert.JSONEq(t, string(invalidJSON), string(results[0]), "should return original bytes unchanged")
}
```

but because the json is invalid we get an fail

### LINKED ISSUE #276: encoded-compare false positive
```
xxx_test.go:80:2: encoded-compare: use assert.JSONEq (testifylint)

  assert.Equal(t, contentTypeJson, header.Header.Get("content-Type"))
```

`const contentTypeJson = "application/json"`, this is not a comparison between two json objects

mmorel-35:
1. Detects only `json.Valid` constants
2. Ignore variables with "negative" words in name
3. Ignore non-strings and non-bytes arguments

Closes #276
Closes #274
Closes #248


ccoVeille:
How this differs from #248 ?

I'm sorry I didn't take time to dig into both to see the differences.

EDIT: I can see you quoted it, so I'll assume it superseded it

PR #302, opened 2026-03-15
feat: add fail-now checker

### LINKED ISSUE #144: fail-now: suggest t.Fatalf instead of weird assertions
Need to find more cases like:

https://github.com/kubernetes/kubernetes/blob/5ec31e84d6c525c173906b1497ee6f075c1926e9/pkg/controller/podautoscaler/horizontal_test.go#L760

mmorel-35:
`assert.Fail`, `assert.FailNow` (and their `f`/`require` variants) are thin wrappers over standard `testing.T` methods. This checker detects their use and auto-fixes to the direct stdlib equivalents.

## Checker: `fail-now`

- **Enabled by default**: true
- **Autofix**: true (when the call result is discarded and the target method exists on the `t` type)

### Fix mapping

| Detected | Fixed |
|---|---|
| `assert.Fail(t, "msg")` | `t.Error("msg")` |
| `assert.Fail(t, "fail", "msg")` | `t.Error("msg")` |
| `assert.Fail(t, "fail", "fmt %s", arg)` | `t.Errorf("fmt %s", arg)` |
| `assert.Failf(t, "fail", "fmt %s", arg)` | `t.Errorf("fmt %s", arg)` |
| `assert.FailNow(t, "msg")` | `t.Fatal("msg")` |
| `assert.FailNow(t, "fail", "fmt %s", arg)` | `t.Fatalf("fmt %s", arg)` |
| `assert.FailNowf(t, "fail", "fmt %s", arg)` | `t.Fatalf("fmt %s", arg)` |
| `s.FailNow("msg")` *(suite receiver)* | `s.T().Fatal("msg")` |

When `msgAndArgs` are present in non-`f` variants, `failureMessage` is dropped and `msgAndArgs` become the replacement arguments (matching testify's own `messageFromMsgAndArgs` convention). Suite method calls are fixed via `receiver.T()` when the receiver implements `TestifySuite`.

### Fix safety constraints

- **Expression context**: the autofix is only emitted when the call result is definitely discarded (`ExprStmt`/`GoStmt`/`DeferStmt`). In expression contexts (e.g. `_ = assert.Fail(...)`, `if assert.Fail(...) {}`), the diagnostic is reported but no fix is suggested, to avoid producing uncompilable code.
- **Method availability**: the fix is gated on the static type of the `t` argument having the target method (checked via `types.LookupFieldOrMethod`). If `t` only implements `assert.TestingT` (which only requires `Errorf`), the non-`f` fix is withheld.
- **Assertion objects**: calls on non-suite method receivers (e.g. `assertObj.Fail(...)` where `assertObj` is `*assert.Assertions`) are ignored entirely — no diagnostic is emitted since there is no accessible `testing.T` to suggest.

## Implementation

- `internal/checkers/fail_now.go` — checker (implemented as `AdvancedChecker` using `insp.WithStack` for parent-node context) + fix logic
- `internal/checkers/checkers_registry.go` — registered in the advanced-checker section, enabled by default
- `internal/testgen/gen_fail_now.go` — test generator with errored + golden templates covering all arg variants, suite method calls, expression-context no-fix cases, and ignored assertion-object calls
- `analyzer/testdata/src/go-require-http-handlers/` — updated test + golden to reflect that `assert.FailNow` and `s.FailNow` in HTTP handlers are now also fixed by this checker

Closes #144


internal/checkers/fail_now.go
```diff
@@ -0,0 +1,190 @@
...
+		switch call.Fn.NameFTrimmed {
+		case "Fail":
+			msg = "use t.Error or t.Errorf instead"
+			proposedFn = "Error"
+		case "FailNow":
+			msg = "use t.Fatal or t.Fatalf instead"
+			proposedFn = "Fatal"
+		default:
+			return true
+		}
```
ccoVeille, line 56: I have a problem here.

testify doesn't expect a *testing.T, but very limited interfaces. One for assert, one for require

https://pkg.go.dev/github.com/stretchr/testify@v1.11.1/assert#TestingT
https://pkg.go.dev/github.com/stretchr/testify@v1.11.1/require#TestingT

The methods might be absent if `t` doesn't implement them.

I would be very careful here. At least with the autofix

ccoVeille (changes requested):

PR #303, opened 2026-03-16
Implement elements-match checker

mmorel-35:
Adds the `elements-match` checker from the [CONTRIBUTING.md open-for-contribution list](https://github.com/Antonboom/testifylint/blob/master/CONTRIBUTING.md#elements-match), which detects verbose slice comparison patterns that should use `assert.ElementsMatch`.

## What it detects

```go
// ❌ Verbose: sort both slices, then compare
slices.Sort(expected)
slices.Sort(result)
assert.True(t, slices.Equal(expected, result))

// ✅ Use the testify idiom
assert.ElementsMatch(t, expected, result)
```

## Changes

- **`internal/checkers/elements_match.go`** — `AdvancedChecker` that walks `ast.BlockStmt` nodes looking for the 3-statement triplet `slices.Sort(x)` + `slices.Sort(y)` + `assert.True(t, slices.Equal(x, y))`. Handles commuted sort order and both `slices` (Go 1.21) and `golang.org/x/exp/slices`. Includes a full autofix that removes the two sort statements and rewrites the assertion.
- **`internal/testgen/gen_elements_match.go`** — Test generator with errored and golden templates.
- **Registry** — Registered as disabled by default (matching the "maybe" stability note in CONTRIBUTING.md).
- **`README.md`** — Added to the checkers table and reference section.
- **`CONTRIBUTING.md`** — Removed from the open-for-contribution list.


ccoVeille:
I had no idea ElementsMatch existed 😱😅🤣

analyzer/testdata/src/vendor/golang.org/x/sys/unix/affinity_linux.go
ccoVeille, line 1: Do you really need to commit all these in vendor?

mmorel-35: Yes ! It is to use "slices"

ccoVeille: But the go.mod is using 1.23 🤔

Why would need exp/slices and why do you talk about Go 1.21 ?

mmorel-35: Slices requires at least Golang 1.23 . I just updated the go mod to 1.25.0. As dependencies are vendored some transitive dependencies has been updated

PR #312, opened 2026-06-01
Add `require-len` checker for fail-fast indexed access guards

mmorel-35:
Adds a dedicated `require-len` checker that enforces fail-fast length guards before indexed collection access. `require-error` is left untouched and remains focused solely on error assertions.

- **New `require-len` checker**
  - Reports `assert.Len*` calls that guard indexed access but should use `require` instead
  - Reports `assert.*` calls that use indexed access without a prior `require.Len` / `require.NotEmpty` guard
  - Provides autofix that inserts `require.Len(t, arr, N)` before the assertion, using `require.NotEmpty(t, arr)` when only index `0` is accessed
  - The `hasLenGuard` check only accepts fail-fast (`require.*`) guards and verifies the asserted length is `>= maxIndex+1`

- **Missing `require` import handled in autofix**
  - When the file does not yet import `github.com/stretchr/testify/require`, the fix also inserts the import
  - Uses `require` as the qualifier; falls back to `require1`, `require2`, etc. if `require` is already taken as a local identifier

- **Registered as enabled-by-default** with dedicated README docs, analyzer fixtures, and golden autofix output


ccoVeille:
While I like the feature it brings, I'm a bit annoyed by the complexity of the code.

I know this comment is not actionable, I'm sorry.

But I wanted to share my point of view

PR #294, opened 2026-03-13
Add `error-compare` checker: detect `err.Error()` in assertions

mmorel-35:
Implements the `error-compare` checker (issue #47). Using `err.Error()` in testify assertions compares error string representations — fragile, ignores the error chain, and bypasses dedicated testify error APIs.

## Detected patterns

| Bad | Suggested fix | Autofix |
|-----|---------------|---------|
| `assert.Contains(t, err.Error(), "substr")` | `assert.ErrorContains(t, err, "substr")` | ✅ |
| `assert.Contains(t, err.Error(), sentinel.Error())` | `assert.ErrorIs(t, err, sentinel)` | ✅ |
| `assert.Equal(t, err.Error(), "msg")` | `assert.EqualError(t, err, "msg")` | ✅ |
| `assert.Equal(t, "msg", err.Error())` | `assert.EqualError(t, err, "msg")` | ✅ (args swapped) |
| `assert.Equal(t, err, sentinel)` *(both implement error)* | `assert.ErrorIs(t, err, sentinel)` | ❌ |
| `assert.NotEqual(t, err, sentinel)` *(both implement error)* | `assert.NotErrorIs(t, err, sentinel)` | ❌ |

All `f`-suffixed variants are handled automatically.

## Implementation

- **`helpers_error.go`**: Added `isErrErrorCall()` (detects `x.Error()` where `x` implements `error`) and `implementsErrorIface()` (type-based, not restricted to `error`-typed variables).
- **`error_compare.go`**: New `RegularChecker` handling `Contains`, `Equal`, and `NotEqual` cases. Guards against overlap with `error-nil` via `isNil` check on `Equal(t, err, nil)`. No autofix is provided for the two-error `Equal`/`NotEqual` cases because `ErrorIs`/`NotErrorIs` are not symmetric and argument order cannot be safely inferred from types alone.
- **Registry**: Inserted between `contains` and `error-nil`, enabled by default.
- **Testgen**: Separate errored and golden templates; `Contains` and `Equal(err.Error(), ...)` cases are autofixed, while `Equal`/`NotEqual` two-error cases remain unchanged in the golden output.
- **README**: Checker documented and added to the summary table.

Also reverts an unrelated change from a previous session that added a "redundant assertion" detection to `require-error`.

Closes #47

PR #319, opened 2026-06-22
Add ErrorContains checker

thespags:
Another case of users using `require.Contains(t, err.Error, ...)` instead of the built in testify `require.ErrorContains(t, err, ...)`


Antonboom:
Duplicate of https://github.com/Antonboom/testifylint/pull/294

thespags:
Thanks @Antonboom that MR is much more complete. Do you know when you'll have a chance to review it?

Antonboom:
I will review and take into accounts all MRs, but in milestones order.

If you have some proofs, that ErrorContains is really useful, I can think to include it in the nearest release

from my PoV, and Contains and ErrorContanis are not proper way to handle error, so we should not suggest any of these

